PR #44, opened 2026-07-21
feat: decision recorder (record.py) + guard subtemplate

### LINKED ISSUE #37: Decision Recorder (`record.py`) — writer-side tooling
## Context
Split from the `decision-memory` ticket (2026-07-16 revision; both tickets refined together in the 2026-07-21 revision): that repo is data-only; all writer tooling lives here in the template repo. This ticket specifies the recorder. The contract it must satisfy (`docs/conventions.md` + CI guards) lives with the data, not here.

Designed dojo-compatible by construction: the record contract ("immutable JSON file, ID = `«timestamp»-«slug»`, minted `v`, envelope `type`, unknown fields tolerated") is the universal write format. decision-memory is a **terminal store** (no compaction; per-record PR review is the point). Future dojo instances reuse the same core and add inbox→JSONL compaction downstream. One writer core, one validator core, two lifecycle policies.

Rulings from the 2026-07-21 grilling session — to be backfilled as the first decision records once this tool exists (see §5).

## Goal
A single-file, stdlib-only `record.py` in the template repo that lets any agent session record, check, and submit decision records to the decision-memory repo with zero manual journaling friction.

---

## 1. Placement & structure
- [ ] Single file: `tools/record.py`, stdlib-only, in the template repo (stamped into projects via Copier). No PyPI package yet — extraction happens when the dojo becomes contract-instance #2
- [ ] **Authoritative behavior doc:** `record.py`'s module docstring, which doubles as `--help` — reviewed and shipped in the same diff as the code, so it cannot drift. NO committed spec file (supersedes the earlier `ticket-decision-recorder.md` plan: a hand-maintained copy of the docstring could only drift). Design history stays in this issue and, once backfilled, in the decision records; the decision-memory ticket points at `--help` and here
- [ ] **Store subtemplate (drift control across stores):** the guard subtemplate vendors not just the CI guard but ALL store-generic files — `docs/conventions.md`, `docs/extraction-prompt.md`, README/AGENTS/CLAUDE, `.gitignore` — plus a seeded, store-owned `preferences.md` (never overwritten on update, conditional `_skip_if_exists`). Multiple stores are planned (one per principal — person or team — so personal and team preferences never mix); N hand-rolled doc copies would silently drift from the schema, while N vendored copies lag visibly and update with one `copier update`, human-gated per store. A schema change lands ONCE here (validator + conventions + extraction prompt in one fixture-tested PR)
- [ ] **Internal seam (dojo lift-target):** strict function boundary between
  - **contract core** (pure, no IO): `mint_id`, `mint_envelope` (`v`, `id`, `type`), `draft_to_record`, `resolve_batch_refs`, `serialize_record` — the functions the dojo package later lifts verbatim; validation is imported from the vendored validator, never duplicated
  - **IO shell**: CLI, clone/branch/commit mechanics, `gh` PR calls. `gh`/PR logic is the one GitHub-specific piece; a hosting supersession swaps this shell function, not the tool. The shell must never ASSUME `gh` is present: managed environments (e.g. Claude Code on the Web) sabotage `gh`/`curl` and declare their own tooling instead (per `CLAUDE.md` → Environment Tooling) — see the `submit` handoff mode below

## 2. Configuration
- [ ] `DECISION_MEMORY_URL` replaces DECISION_MEMORY_REPO. env var: **full git URL** of the decision-memory repo (preserves hosting optionality). Never committed anywhere public
- [ ] Auth: whatever git/`gh` credentials the agent account already has; the tool adds no credential handling of its own

### `open`
- [ ] Shallow-clone `DECISION_MEMORY_URL` into an ephemeral temp dir (per session; no persistent clone — stateless across machines)
- [ ] **Attached-checkout reuse:** where cloning is impossible (managed environments may block outbound clones while the store repo is already attached to the session), `open --use «path»` validates the checkout — origin matched against `DECISION_MEMORY_URL` by owner/repo tail, since proxy-rewritten remotes never match textually; clean worktree required — and runs the normal session flow in place. A failed clone points at the flag
- [ ] Capture HEAD SHA → this session's `preference_set.commit`
- [ ] Create branch `session/«YYYYMMDDTHHMMSSZ»`
- [ ] **Unmerged-PR sweep (stateless):** list ALL closed-unmerged PRs of the decision-memory repo, subtract those already covered by an existing record with a matching `closure_of`, and prompt one decision record per remaining closure ("why was PR #N rejected"), `correction` flag where applicable; the new record carries `closure_of: «pr-number»`. The records themselves are the sweep watermark — no timestamps, no local state, idempotent across machines, self-healing (a closure skipped in one session resurfaces in the next). Mechanism for the ticketed "closure is itself signal" — no new tooling, reuses the record format
- [ ] Print `preferences.md` injection reminder + captured SHA

### `record`
- [ ] Mint `id` (UTC timestamp + writer-chosen kebab slug ≤40 chars), `v`, `date`, `session`, `type: "decision"`
- [ ] Validate: envelope + decision schema + field conventions (exactly one prediction-role option; `rules_cited` non-empty iff `prediction_stream` == preference-driven; `operative_reason` rules incl. silent picks via `operative_reason_source`; rejection `reason_source` provenance — null reasons only under declared `none`; recommendations never back-filled — input side must be supplied before the ruling is)
- [ ] Write `decisions/«id».json`; ONE commit: `decision(«project»): «slug» — «chosen»`
- [ ] **Batch ingest:** `record --from «drafts.json»` accepts an array of **draft records** — the record schema minus tool-minted fields, as produced by the data repo's `docs/extraction-prompt.md` (data-repo §4) — and runs the normal single-record path per entry (mint, validate, write, one commit per record). Convenience only: looping single `record` calls is equivalent
- [ ] **Batch-local references:** drafts may name other drafts in the same batch by slug — `supersedes_slug` / `drill_down_of_slug` (single), `related_slugs` (list, merged with repo-ID `related` entries); ingestion resolves all of them to minted IDs (order-independent; unknown slugs and slug/resolved-field conflicts fail loudly). Replaces prose cross-references in notes

### `check`
- [ ] Validate ALL of `decisions/` + dangling `related`/`supersedes`/`drill_down_of` refs + `preferences.md` token budget
- [ ] **Single validator source (supersedes B3):** one stdlib-only validator file lives HERE in the template repo; `record.py` imports it directly (no hand-duplication), decision-memory vendors the same file via a minimal guard subtemplate + `copier update`. Permanent mechanism — PyPI/runtime-install plan dropped (no registry dependency; repo stays self-contained, fails soft on factory loss: guard keeps working, only updates stop). Updates land as reviewable diffs in the data repo — human gate on guard changes preserved. Drift = vendored copy lagging; surfaces as rejected PRs, fixed by `copier update`
- [ ] Fixture tests for schema edge cases live here beside the validator; every validator change must pass them in this repo's CI before it can be vendored downstream

### `submit`
- [ ] Compute per-session prediction hit rates as two streams (preference-driven vs cold); `refined` (chosen contains the prediction plus an extension; requires a prediction-slot mismatch — same slot is a plain hit) and near-tie outcomes bucketed separately — never counted as misses
- [ ] **Auto-bump:** for each CLEAN preference-driven hit (`refined` never bumps), generate the `pref-confirm: «rule» (n=«count»)` counter-line commit — provenance is already in the record (`rules_cited` + `chosen_slot` + `outcome`); CI validates the counter math. Self-confirmation caveat stays extraction-side per data-repo §5
- [ ] Open PR via `gh`: description = two-stream hit rates + any `supersedes` claims surfaced for explicit review
- [ ] **Managed-environment handoff mode:** `record.py` is a Python script and cannot call environment-declared tooling (e.g. GitHub MCP) itself. Where `gh` is unavailable/sabotaged, `submit` does everything up to the PR — commits, branch push, computed two-stream hit rates — then emits the PR title + body and stops; the driving agent opens the PR with the tooling its environment declares (per `CLAUDE.md` → Environment Tooling). Same seam as the hosting-supersession swap: the PR call is shell, not core
- [ ] One PR per session; one commit per record (partial accept = human edits the branch)

### `propose`
- [ ] Write `proposals/«YYYY-MM-DD»-«slug».md`, one rule per file, commit `pref-proposal: «rule»`
- [ ] Thin verb on purpose: enforces file/commit convention at write time instead of by CI rejection. Promotion stays human-only (`pref-promote`)

## 4. Grilling-skill integration
- [ ] The grilling skill calls the verbs; it never writes records or commits directly
- [ ] Slot model maps 1:1 onto the record schema (slot 1 prediction, slot 2 recommendation, slot 3 wildcard, slot 4 free choice). Free-text answers like "Option N, because XYZ" upgrade the presumed-false rejection reason to operative; silent MC taps of a non-prediction option record `operative_reason_source: none`
- [ ] Session flow: skill runs `open` at start (SHA capture + unmerged-PR sweep), `record` per ruling, `check` then `submit` at end

## 5. First run (backfill — first beta test)
The rulings behind this ticket and the decision-memory ticket were made in the grilling conversations that produced them. Backfilling them is the tool's first beta test: run the data repo's `docs/extraction-prompt.md` over the originating conversations to produce draft records, then ingest them USING the recorder itself (`open` → `record` per draft or `record --from` → `check` → `submit`), not by hand-authoring files.

**Status:** TWO extraction beta batches exist and both now verify clean end-to-end against the final schema. Batch 1 (24 records, the originating grilling conversations — updated in its origin session to the refined schema). Batch 2 (9 records, a live disambiguate grilling session, run to verify the final prompt) — it surfaced and drove the silent-pick `operative_reason_source`, the `refined` slot-mismatch rule, generalized batch-local slug references, and the partial-artifact-ref/chosen-verbatim/if-clause-literal prompt refinements; its two silent picks need the two-line `operative_reason_source: "none"` declaration at ingestion (mechanical, no re-extraction).

- [ ] **Batch updates happen in the ORIGINATING session** — it alone holds the conversation context needed to infer reasons; the ingestion session must never guess from labels (never-invent rule). Ingestion-side fixes stay mechanical: `artifact_ref` enrichment (incl. filling `commit: null` in partial refs), declared-silent annotations, format repairs, ID resolution
- [ ] Ingest batch 1 as the first `session/«ts»` PR (then batch 2). Known rulings covered by batch 1: A1, A2, B1, B2, B3, C1, C2, C3, D3, D5, D6, plus the dojo-side and sequencing rulings and the same-day follow-ups (copier-vendoring incl. permanence, chat-draft extraction)
- [ ] Additional rulings made SINCE the extraction (this session) get their own records: no-spec-file (`--help` authoritative), store subtemplate, attached-checkout reuse, no-duplication convention, prose/code placeholder scoping, the `reason_source`/`operative_reason_source`/`refined`/batch-refs schema evolution, and a disposition record for the stale pre-unification decision-memory implementation (discarded — encodes overruled rulings in an incompatible schema without supersedes chains)

## Non-goals
- No PyPI package, no separate repo — single file until the dojo provides the second contract instance. (Contract-core lift into the dojo package stands unchanged; VALIDATOR distribution stays copier-vendored permanently, per the B3 supersession)
- No persistent local clone, no sync state
- No retrieval/RAG, no history injection — consumers inject `preferences.md` only (data-repo §8)
- No automatic preference promotion — `pref-promote` stays human
- No dojo compaction logic here — that ships with the dojo, consuming the same contract core
- No MCP/forge-API client code in `record.py` — environment-tooling PR creation is the agent's job via the handoff mode, never the tool's
- No chat-side extraction tooling — extracting records from past conversations is a PROMPT (`docs/extraction-prompt.md`, vendored into each store) emitting draft records; validation happens only at ingestion, where the single vendored validator lives
- No committed spec file and no ADR for the recorder — behavior lives in `--help`, the data contract in the vendored `docs/conventions.md` + validator, rationale in the backfilled decision records (one home per layer)
- No per-store policy parameterization (budgets, commit types) until a second store actually demands it — no speculative design
- No `--patch` flag for artifact-ref injection — drafts are plain JSON, edited freely before ingestion; revisit only if manual editing proves to be friction
- No inferred tier for `operative_reason` — operative means decider-confirmed; inference lives in option `reasoning` and rejection reasons only

frankify-app:
Closes #37

## What's included

**Store subtemplate (new `guard/` subtemplate, selected via `agentic_subtemplate`):** the copier template now hosts a second subtemplate using the documented templated-`_subdirectory` pattern. `agentic_subtemplate=guard` stamps a complete decision store: the CI guard (`decision_validator.py` — the single validation source; `guards.py` — append-only checks, full-corpus schema + consistency validation, dangling refs, preferences token budget, pref-confirm counter math, repo-type commit lint; the `guards.yml` workflow) AND the vendored store docs (`docs/conventions.md`, `docs/extraction-prompt.md`, README/AGENTS/CLAUDE, `.gitignore`) plus a seeded, store-owned `preferences.md` (conditional `_skip_if_exists`, never overwritten on update). Docs travel with the schema so N stores (one per principal — person or team) cannot silently drift from it: a schema change lands once here and reaches every store via `copier update`, human-gated per store. All project-scaffold questions and post-copy tasks are gated on the default `template` choice, so existing consumers see no change and a store's answers file stays minimal.

**`tools/record.py` (stamped into projects, self-applied at root):** single-file stdlib-only recorder with a strict contract-core / IO-shell seam. Verbs: `open` (ephemeral shallow clone — or `--use «path»` to reuse an already-attached checkout where cloning is blocked, matched against `DECISION_MEMORY_URL` by owner/repo tail since proxy-rewritten remotes never match textually; SHA capture, session branch, stateless closed-unmerged-PR sweep via `closure_of`), `record` (stdin draft or `--from drafts.json` batch ingest, one commit per record), `check` (full-corpus validation), `submit` (two-stream hit rates, mechanical `pref-confirm` bumps, push, `gh` PR with managed-environment handoff fallback), `propose`.

**No committed spec file:** `record.py`'s module docstring (= `--help`) is the authoritative behavior doc — reviewed in the same diff as the code, so it cannot drift. The data contract is the vendored `docs/conventions.md` + validator; design rationale lands as backfilled decision records. One home per layer.

**`DECISION_MEMORY_URL` (breaking):** replaces `DECISION_MEMORY_REPO`; the variable now carries the full git URL. AGENTS.md, doctor.sh, and tests updated.

**Tests:** validator fixtures, guard pure functions, subtemplate render (exact file set, byte-identical vendoring, seeded-preferences survival on re-render), recorder contract core, URL-tail matching. 81 passed locally; 2 environmental failures are the host-tool smoke tests needing `gh` (absent in the dev sandbox; CI has it); all prek hooks pass except shellcheck, whose binary download the sandbox proxy blocks — CI runs it.

## Obstacles / notes

- Local test runs were silently rendering the latest **release tag** instead of the working branch (release tags exist locally; CI checkouts have none and fall back to HEAD). Fixed by pinning `vcs_ref="HEAD"` in the test render helpers — local runs now match CI.
- Validation deliberately lives in ONE file (`guard/.github/guards/decision_validator.py`); `record.py` imports the copier-vendored copy from the store checkout at runtime.
- `open --use` was verified live against a proxy-rewritten checkout (CCoW-style origin vs GitHub `DECISION_MEMORY_URL`).

## DECISION markers

- `DECISION:` (record.py, contract core) — the validator is imported from the store checkout's vendored copy (single copier-vendored source shared with CI), so writer-side and CI validation cannot drift.
- `DECISION:` (record.py, `open`) — session state lives inside the ephemeral clone (untracked file, excluded from git status); nothing persists outside the temp dir, keeping sessions stateless across machines.
- `DECISION:` (record.py, PR listing) — any `gh` failure falls through to the managed-environment handoff path; in sabotaged environments failure is an expected mode, not an error.

https://claude.ai/code/session_014dUrs8wH2HRRB4JG58bVQR

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -81,7 +81,7 @@ Code-specific skills:
 
 ### Skill Environment Variables
 
-- `DECISION_MEMORY_REPO` — URL of the decision-memory repo the `grilling` skill records decisions to. Recording requires this env var in the agent's execution environment; the skill reads exactly this name (shared contract with the skill — renaming either side breaks recording silently). Never hardcode, commit, or echo the value into artifacts. Unset → grilling still works, skips recording, and says so. Where to set it: local sessions → shell profile / user-level agent settings; remote or cloud sessions → the environment's configuration; CI → a repository secret. `scripts/doctor.sh` warns when it's unset and checks reachability when set.
+- `DECISION_MEMORY_URL` — FULL git URL of the decision-memory repo that `tools/record.py` (and the `grilling` skill through it) records decisions to. A full URL rather than an owner/repo slug, so the hosting stays swappable. Recording requires this env var in the agent's execution environment; the tool and the skill read exactly this name (shared contract — renaming either side breaks recording silently). Never hardcode, commit, or echo the value into artifacts. Unset → grilling still works, skips recording, and says so. Where to set it: local sessions → shell profile / user-level agent settings; remote or cloud sessions → the environment's configuration; CI → a repository secret. `scripts/doctor.sh` warns when it's unset and checks reachability when set.
 
 ## Git
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -107,6 +107,10 @@ Placeholder syntax:
   `«` and `»` — e.g. `decisions/«id».json`, `«timestamp»-«slug»`.
   Never `<placeholder>` — it is stripped on programmatic reads and
   edits.
+- Guillemets are for tracker-posted content ONLY. In code (source
+  files, comments, error messages, help text) use plain
+  `<angle-bracket>` placeholders — committed files never pass through
+  a tracker sanitizer, and guillemets are alien there.
 
 Line wrapping:
 
--- a/copier.yml
+++ b/copier.yml
@@ -1,18 +1,35 @@
 # questions
+agentic_subtemplate:
+  type: str
+  help: >-
+    Which subtemplate to render: `template` is the full project scaffold;
+    `guard` stamps a decision-memory STORE — the CI guard (validator +
+    guard runner + workflow), the vendored store docs (conventions,
+    extraction prompt, README/AGENTS/CLAUDE), and a seeded preference
+    set. With `guard`, all project-scaffold questions are skipped,
+    keeping the store's answers file minimal.
+  choices:
+    - template
+    - guard
+  default: template
+
 agentic_project_name:
   type: str
   help: "Human-readable project name (agent docs, glossary)."
   placeholder: "My Project"
+  when: "{{ agentic_subtemplate == 'template' }}"
 
 agentic_project_description:
   type: str
   help: "Short project description (glossary seed entry)."
   placeholder: "A helpful project."
+  when: "{{ agentic_subtemplate == 'template' }}"
 
 agentic_project_slug:
   type: str
   help: "Project slug in kebab-case (repo URL, glossary path). Auto-derived from name; override if needed."
   default: "{{ agentic_project_name | lower | replace(' ', '-') | replace('_', '-') }}"
+  when: "{{ agentic_subtemplate == 'template' }}"
 
 agentic_project_kind:
   type: str
@@ -21,16 +38,19 @@ agentic_project_kind:
     - code
     - docs
   default: code
+  when: "{{ agentic_subtemplate == 'template' }}"
 
 agentic_language:
   type: str
   help: "Primary content language as an ISO 639-1 code (e.g. en, de). Language-sensitive tooling (codespell) is only rendered for English content."
   default: en
+  when: "{{ agentic_subtemplate == 'template' }}"
 
 agentic_disambiguate_version:
   type: str
   help: "Pinned disambiguate version (pre-alpha — breaking changes between releases; hook and docs commands use this pin)."
   default: "0.2.0"
+  when: "{{ agentic_subtemplate == 'template' }}"
 
 agentic_disambiguate_roots:
   type: str
@@ -40,6 +60,7 @@ agentic_disambiguate_roots:
     Empty keeps disambiguate's default roots. Kept as answer data so repos
     never patch the template-owned .pre-commit-config.yaml.
   default: ""
+  when: "{{ agentic_subtemplate == 'template' }}"
 
 agentic_precommit:
   type: str
@@ -48,6 +69,7 @@ agentic_precommit:
     - prek
     - none
   default: prek
+  when: "{{ agentic_subtemplate == 'template' }}"
 
 agentic_tracker_cli:
   type: str
@@ -57,6 +79,7 @@ agentic_tracker_cli:
     - gh
     - tea
   default: ghx
+  when: "{{ agentic_subtemplate == 'template' }}"
 
 agentic_forge:
   type: str
@@ -71,7 +94,7 @@ agentic_forge:
     {%- else -%}
     github
     {%- endif %}
-  when: "{{ not detect_forge() }}"
+  when: "{{ agentic_subtemplate == 'template' and not detect_forge() }}"
 
 agentic_forgejo_host:
   type: str
@@ -99,11 +122,12 @@ agentic_repo_owner:
     {%- else -%}
     {{ UNSET }}
     {%- endif %}
-  validator: "{% if not agentic_repo_owner %}Could not detect repo owner — run `gh auth login` or `git config github.user <login>`, or enter manually{% endif %}"
+  validator: "{% if agentic_subtemplate == 'template' and not agentic_repo_owner %}Could not detect repo owner — run `gh auth login` or `git config github.user <login>`, or enter manually{% endif %}"
+  when: "{{ agentic_subtemplate == 'template' }}"
 
 # Copier metadata
 _min_copier_version: "9.11.0"
-_subdirectory: template
+_subdirectory: "{{ agentic_subtemplate }}"
 # Keep the .claude/skills → ../.agents/skills bridge a symlink in generated
 # repos instead of copying (a copy would be empty — skills install post-render).
 _preserve_symlinks: true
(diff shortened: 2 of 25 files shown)


frankify-app:
Re: clone availability in managed environments — implemented as `open --use «path»`: it validates an already-attached checkout instead of cloning. Matching is by owner/repo tail of the origin URL rather than full-URL comparison, because managed environments rewrite remotes through local proxies (e.g. `http://local_proxy@127.0.0.1:«port»/git/«owner»/«repo»`), so the origin never matches `DECISION_MEMORY_URL` textually. A clean worktree is required, and a failed clone now points at the flag. Verified live against a proxy-rewritten checkout.

I stopped short of auto-discovery (scanning the filesystem for candidate repos): the driving agent already knows which repos its session has attached, so an explicit path is one flag with no magic, no false positives. Can revisit if the manual step turns out to be friction.

---

hvoecking (dismissed):

Issue #60, opened 2026-07-25
copier `_tasks` doctor.sh call aborts stamp/update when a host tool is missing


frankify-app:
## Problem

`copier.yml`'s `_tasks` runs `bash scripts/doctor.sh`. Doctor exits non-zero when a **required** host tool is missing, and a non-zero task aborts the whole copier run — copier rolls the update back entirely:

```console
$ copier update --trust --defaults --answers-file .copier-answers.agentic.yml --vcs-ref v2.0.1
...
Re-run with --install to bootstrap missing host tools.
brew install gh  OR  sudo apt-get install -y gh
Task 'bash scripts/doctor.sh' returned non-zero exit status 1.
# → working tree unchanged, update lost
```

The missing tool here is `gh` — which is **deliberately absent** in Claude Code remote sessions, where GitHub work goes through MCP tools (the repo CLAUDE.md files say exactly this: "Never fall back to `gh`, `ghx`, `curl` … the environment actively sabotages them"). So the template's own task list contradicts the template's own environment guidance, and every `copier update` in a remote agent session fails unless the operator knows to pass `--skip-tasks`.

Observed on frankify-app/skills while working frankify-app/skills#10; the same trap applies to any generated repo updated from an environment without the full required-tool set.

## Options

1. **Make the doctor task advisory** — `bash scripts/doctor.sh || true` in `_tasks` (report, never abort). Preferred: doctor's value is the report; blocking a scaffold on a missing host CLI is a heavy failure mode for a check that cannot fix itself without `--install`.
2. **Demote `gh` from required to warn** when a tracker CLI is configured/MCP is in use — narrower, but leaves other tools able to abort updates.
3. Keep failing, and document `--skip-tasks` prominently in the template-update workflow and AGENTS.md.

Recommend 1 (plus keeping doctor's own exit code meaningful when run directly, so CI/humans still get a hard signal).

## Acceptance

- [ ] `copier copy`/`update` completes in an environment missing `gh`, with the doctor report still shown.
- [ ] Directly-invoked `scripts/doctor.sh` still exits non-zero on missing required tools.
- [ ] Template-update docs state which flags a remote agent session needs, if any.

Issue #57, opened 2026-07-25
skills-lock computedHash is not verified at install — pins don't pin, drift lands silently


frankify-app:
## Problem

Observed while adopting #54 on disambiguate: `npx skills@latest experimental_install` completed successfully even though `skills-lock.json`'s `computedHash` for `tdd` predates the current content of `frankify-app/skills` — and the install **modified two already-vendored skills** (`grilling`, `requesting-code-review`) to newer upstream content **without touching the lock file**. No warning, no failure.

So the lock currently documents a past state rather than enforcing one:

- `computedHash` is not validated against fetched content (or mismatches are silently accepted).
- Vendored `.claude/skills` / `.agents/skills` content can drift from the lock on any install, defeating the point of pinning and making skill provenance unauditable.
- Also unclear/undocumented: what `computedHash` is actually computed over (it is not a plain sha256 of `SKILL.md`), which blocks refreshing pins by hand — concretely, the `tdd` pin can't be updated for frankify-app/skills#13 without reverse-engineering the CLI.

## Proposal

1. Decide the semantics: either the lock **pins** (install fails/warns loudly on hash mismatch, explicit `update` command refreshes lock + content together) or it is a **manifest** (then rename/document it as such and stop implying immutability).
2. Document the hash algorithm in `docs/conventions.md` (or wherever the lock contract lives) so locks can be audited and refreshed deliberately.
3. Template-update flow should include a lock-refresh step when upstream skills change (concrete pending case: frankify-app/skills#13).

If the `skills` CLI itself is the limiting factor, this issue is the template-side tracking for working around or upstreaming it.

frankify-app:
Data point refining the original report: on a re-run of `npx skills@latest experimental_install` after frankify-app/skills#13 merged, the installer **did** rewrite `skills-lock.json` (refreshed `computedHash` for the four changed skills alongside the content). So current semantics are confirmed as **manifest-that-follows-content**: install fetches latest upstream, updates vendored files, and refreshes the lock to match — it never gates on the recorded hash (the first observation in this issue, where content changed without a lock rewrite, remains unexplained but was likely a partial-write state). The decision in this ticket stands: either make it a real pin (fail on mismatch, explicit update command) or rename/document it as a manifest — but nothing today prevents an unreviewed upstream skill change from landing in a downstream repo on any install.

---
_Generated by [Claude Code](https://claude.ai/code)_

frankify-app:
[...]

PR #62, opened 2026-07-25
feat: source-trust policy for vendored skills

### LINKED ISSUE #58: Source-trust policy for vendored skills: auto-update trusted sources, human-gate the rest
Implements the decision recorded on #57 (read that ticket first — it establishes that `skills-lock.json` is a manifest that follows content, never a gate).

## Problem

Vendored skills are executable instructions running with full agent permissions (the installer says so on every run). Today `npx skills experimental_install` fetches latest from **every** source and rewrites both the vendored files and the lock — so an unreviewed upstream change from a third-party repo lands in every future agent session of every generated repo, silently. Pinning everything is the safe default but blocks wanted updates from repos we control; auto-updating everything is a remote-write channel into our agents.

## Policy model

Trust is **per source repo**, not per skill:

- **Trusted source** = every change to it already passed our review gate before landing (currently `frankify-app/skills`: human-merged PRs). Auto-updating from it does not skip review — the review happened upstream.
- **Untrusted source** = everything else (`mattpocock/skills`, `obra/superpowers`, …): changes nobody in the org has read. Pin, and route updates through human review.

## Scope

1. **`skills-policy.json`** (new, template-owned, rendered into generated repos) — the CLI must not own this file, since it rewrites `skills-lock.json` on every install:

   ```json
   { "trustedSources": ["frankify-app/skills"] }
   ```

   Seed the trusted list from a copier answer so downstream repos can extend it.

2. **`scripts/update-skills.sh`** (new) — the only sanctioned way to run the installer:
   - snapshot vendored skill content (`.agents/skills/`) + lock;
   - run `npx skills@latest experimental_install`;
   - per skill, diff content against the snapshot and resolve its source via `skills-lock.json`;
   - source trusted → keep the change; source untrusted → **revert** content *and* the lock entry to the snapshot, and append the diff to a review report (`skills-review.md`, gitignored or PR-body material).
   - Diff on **content**, not `computedHash` — the hash is opaque and gets refreshed to match whatever was fetched (#57), so it carries no integrity signal.
   - First-time vendoring of a new skill from an untrusted source goes through the same quarantine path — initial adoption is the highest-risk moment, not just updates.
   - Exit non-zero when untrusted drift was quarantined, so CI can branch on it.

3. **`.github/workflows/skills-update.yml`** (new, scheduled + manual) — runs the wrapper:
   - trusted drift → PR labeled `skills-auto`, allowed to auto-merge on green CI (auto-merge rather than direct commit, so there is a CI record and a revert point even for the trusted lane);
   - untrusted drift → separate PR labeled `needs-human-review`, never auto-merged, full `SKILL.md` diff inline in the body plus the injection-shaped review checklist (new instructions that escalate access, touch credentials, call external services, or countermand other rules).

4. **CODEOWNERS render** — vendored paths of untrusted skills get a CODEOWNERS entry so branch protection requires human approval mechanically, not just by convention. (If per-skill paths are too fine-grained, own `.agents/skills/**` and let the trusted lane's auto-merge be the exception.)

5. **Docs** — `AGENTS.md` (or `docs/conventions.md`): never call `npx skills … install` directly; always `scripts/update-skills.sh`. Record the trust model and how a source gets promoted to trusted.

## Acceptance

- [ ] With `frankify-app/skills` trusted and a change present upstream in it, the wrapper keeps the change and exits 0.
- [ ] With a change present in an untrusted source, the wrapper reverts content + lock entry, writes the review report, exits non-zero.
- [ ] A brand-new skill from an untrusted source is quarantined, not silently vendored.
- [ ] Scheduled workflow produces the two PR lanes with the right labels; the untrusted lane cannot auto-merge.
- [ ] Wrapper is idempotent: a second run with no upstream change is a no-op, exit 0.

## Residual risk (accepted, documented)

Compromise of a trusted source auto-lands. Mitigated by auto-merge-on-green (CI record + revert point) rather than direct commit, and by the trusted list staying deliberately short.

## Non-goals

Patching the upstream `skills` CLI — see #59 for the pin-mode request; this ticket must work with the CLI as-is.

frankify-app:
Closes #58. Implements the ruling recorded on #57: the lock stays a manifest, and trust moves to the source level.

## The problem

`npx skills … install` fetches latest from **every** source and rewrites both the vendored files and the lock. Since vendored skills are executable instructions running with full agent permissions, that is a remote-write channel into every future agent session — an unreviewed upstream change from a third-party repo lands silently.

## What ships

- **`skills-policy.json`** — `trustedSources` allowlist, seeded from a new `agentic_trusted_skill_sources` copier answer (default `frankify-app/skills`).
- **`scripts/update-skills.py`** — the only sanctioned way to install. Snapshots the vendored tree, runs the installer, then for every skill from a source outside the allowlist: reverts the content **and** the lock entry, and writes the diff to `skills-review.md`. Exit 0 = clean, 1 = quarantined, 2 = installer failed.
- **`.github/workflows/skills-update.yml`** — two lanes. Trusted drift opens a PR labeled `skills-auto` with `gh pr merge --auto`; untrusted drift opens a `needs-human-review` PR that never auto-merges.
- **`.github/CODEOWNERS`** — owns `.agents/skills/` and `skills-policy.json`, so with branch protection the untrusted lane's human gate is mechanical, not conventional.
- **AGENTS.md § Vendored Skills** — never call the installer directly; trust is per source; reviewing a quarantined skill is a *security* review (instructions that escalate access, touch credentials, reach external services, or countermand other rules).

## Two design points worth review

**Auto-merge on the trusted lane.** Confirmed with @frankify-app: every change to `frankify-app/skills` already lands via a human-reviewed PR, so the downstream PR carries bytes that were already approved — a second review adds latency, not information. Auto-merge-on-green rather than a direct commit keeps a CI record and a revert point. Residual risk (compromise of a trusted source auto-lands) is accepted and documented; the allowlist stays deliberately short.

**Drift is judged on content, never `computedHash`.** The installer rewrites that hash to match whatever it just fetched — confirmed empirically in this session — so it carries no integrity signal. Diffing content is the only thing that actually detects change.

## Verification

TDD, red-first: the five render tests were watched failing before any of the files existed.

Rendering the right files proves nothing about whether the guard *holds*, so there are also four behavior tests that import the wrapper and fake the installer:

| Scenario | Asserted |
| --- | --- |
| Trusted source changes a skill | change **kept**, exit 0, no report |
| Untrusted source changes a skill | content **reverted** to vendored state, exit 1, diff in report (the test plants `ignore all previous instructions` and asserts it is reviewable) |
| Untrusted source adds a **new** skill | directory removed **and** lock entry reverted — first-time vendoring is the highest-risk moment, not just updates |
| Installer itself fails | exit 2, distinguishable from quarantine |

Suite: **106 passed**. The 2 `test_e2e` failures are pre-existing #60 (copier's `_tasks` doctor call hard-failing on missing `gh`) — identical on unmodified `main`, now visible because prek is installed so they no longer skip. `prek run --all-files` clean on every commit.

## Notes

- I dropped a `--check` flag I had written: it snapshotted and compared the same state, so it always reported "no untrusted drift". A flag that cannot do what its help text says is worse than no flag.
- Self-application adopts `AGENTS.md`, `skills-policy.json`, `scripts/update-skills.py`. **Not** the workflow or CODEOWNERS: this repo root carries neither the rendered `lint.yml` nor `template-update.yml` (it has its own `ci.yml`), and it vendors no skills under `.agents/`, so a CODEOWNERS rule there would guard nothing. Say the word if you want the template repo itself on the scheduled update lane.
- Consumers need the `skills-auto` and `needs-human-review` labels to exist, and branch protection with CODEOWNERS review required for the gate to bite. Neither is created by this PR.

https://claude.ai/code/session_019Y1kVb7zPiUcZNnFQzfTLK

---
_Generated by [Claude Code](https://claude.ai/code/session_019Y1kVb7zPiUcZNnFQzfTLK)_

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -256,6 +256,15 @@ Add packages using the package manager only, never edit requirements/dependencie
 - Test cases must cover edge cases for inputs and every @returns line in the contract
 - Non-trivial decisions or behavior should be documented via inline comments
 
+## Vendored Skills
+
+Skills under `.agents/skills/` are executable instructions that run with full agent permissions, so an update is a remote-write channel into every future session.
+
+- **Never run the installer directly** (`npx skills … experimental_install`). Always `python3 scripts/update-skills.py`, which reverts and reports changes from sources outside `skills-policy.json`'s `trustedSources`.
+- **Trust is per source repo, not per skill.** A source belongs in `trustedSources` only when every change to it already passes a review gate you control — then auto-updating skips no review, it inherits one.
+- **Reviewing a quarantined skill is a security review, not a style review**: look for instructions that escalate access, touch credentials, reach external services, or countermand rules elsewhere in this file.
+- Drift is judged on **content**, never on `skills-lock.json`'s `computedHash` — the installer rewrites that hash to match whatever it just fetched, so it carries no integrity signal.
+
 ## Failures Become Rules
 
 When something fails that automation or an instruction could have prevented — a lint run nobody made, a tool nobody installed, a convention discovered only in review — the fix is incomplete until the prevention is encoded: a hook, a doctor check, or a rule in the template repo's AGENTS.md/skills (preferred, so every generated repo inherits it); [docs/conventions.md](docs/conventions.md) only when it is genuinely repo-local. File the ticket on the owning repo in the same session the failure surfaced. Fixing only the instance guarantees a repeat.
--- a/copier.yml
+++ b/copier.yml
@@ -62,6 +62,17 @@ agentic_disambiguate_roots:
   default: ""
   when: "{{ agentic_subtemplate == 'template' }}"
 
+agentic_trusted_skill_sources:
+  type: str
+  help: >-
+    Comma-separated skill sources (owner/repo) allowed to AUTO-UPDATE into
+    this repo. Vendored skills are executable instructions with full agent
+    permissions, so list a source only when every change to it already
+    passes a review gate you control. Everything else is quarantined by
+    scripts/update-skills.py for human review.
+  default: "frankify-app/skills"
+  when: "{{ agentic_subtemplate == 'template' }}"
+
 agentic_precommit:
   type: str
   help: "Install cross-cutting prek hooks (commitlint, JSON, Markdown, codespell)?"
--- a/scripts/update-skills.py
+++ b/scripts/update-skills.py
@@ -1,0 +1,192 @@
+#!/usr/bin/env python3
+"""
+Guarded skills install: auto-update trusted sources, quarantine the rest.
+
+Vendored skills are executable instructions that run with full agent
+permissions, so `skills … install` — which fetches latest from EVERY
+source — is a remote-write channel into every future agent session.
+
+Trust is a property of the SOURCE repo, not of a hash: a source whose
+changes already pass a human review gate upstream (our own skills repo)
+can auto-update, because the review already happened. Everything else is
+held at the vendored state until a human reads the diff.
+
+Drift is detected by comparing CONTENT, never `computedHash`: the lock is
+a manifest that the installer rewrites to match whatever it just fetched,
+so its hash carries no integrity signal.
+
+Exit codes: 0 = nothing quarantined; 1 = untrusted drift was reverted and
+written to the review report; 2 = the install itself failed.
+"""
+
+from __future__ import annotations
+
+import difflib
+import json
+from pathlib import Path
+import shutil
+import subprocess
+import sys
+import tempfile
+
+POLICY_FILE = Path("skills-policy.json")
+LOCK_FILE = Path("skills-lock.json")
+SKILLS_DIR = Path(".agents/skills")
+REPORT_FILE = Path("skills-review.md")
+INSTALL_CMD = ["npx", "--yes", "skills@latest", "experimental_install"]
+
+
+def trusted_sources() -> set[str]:
+    """Read the trusted-source allowlist; absent policy trusts nothing."""
+    if not POLICY_FILE.exists():
+        return set()
+    policy = json.loads(POLICY_FILE.read_text(encoding="utf-8"))
+    return set(policy.get("trustedSources", []))
+
+
+def read_lock() -> dict:
+    """Return the lock as a dict; a missing lock reads as empty."""
+    if not LOCK_FILE.exists():
+        return {"skills": {}}
+    return json.loads(LOCK_FILE.read_text(encoding="utf-8"))
+
+
+def skill_files(root: Path) -> dict[str, str]:
+    """Map relative path -> text for every file under `root`."""
+    if not root.is_dir():
+        return {}
+    return {
+        str(path.relative_to(root)): path.read_text(encoding="utf-8", errors="replace")
+        for path in sorted(root.rglob("*"))
+        if path.is_file()
+    }
+
+
+def content_diff(before: Path, after: Path, name: str) -> str:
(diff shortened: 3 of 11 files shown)


pando-genet:
Confirming evidence for this PR's position that **drift must be judged on content, never `computedHash`** — now with a real instance rather than a reasoned expectation.

All four factory repos record the identical `computedHash` for the `grilling` skill:

```
82c8da67d4ec62ded3a4b3823b0e4678d93a0acfc22ac41180cb1b4a5f336872
```

| Repo | lock hash | vendored content |
| --- | --- | --- |
| `ghx` | `82c8da67…` | **12 lines — the upstream `grill-me` body** |
| `meta` | `82c8da67…` | 80 lines — the org's `grilling` skill |
| `disambiguate` | `82c8da67…` | 80 lines — the org's `grilling` skill |
| `skills` | `82c8da67…` | 80 lines — source |

Same hash, different bytes. The lockfile asserts an integrity property it does not have, exactly as this PR predicted:

> the installer rewrites that hash to match whatever it just fetched — confirmed empirically in this session — so it carries no integrity signal. Diffing content is the only thing that actually detects change.

Two things sharpen the case beyond the abstract argument:

**The wrong content came from a third-party source.** `ghx`'s `grilling` slot holds text from `mattpocock/skills` — an *untrusted* source under this PR's own allowlist, sitting in a slot whose lock entry names `frankify-app/skills` as its origin. The quarantine lane this PR builds is precisely the control for that, and the hash would not have triggered it.

**It failed silently and consequentially.** The substituted file has no `## Recording` section, so a long planning session run against that repo produced fourteen rulings and recorded none of them, with nothing in the transcript indicating a problem. A wrong skill does not announce itself; it just quietly does less.

Filed as pandoscope/ghx#7. Not asking for a change to this PR — the design already handles it. Recording it here so the rationale has an instance attached, since a maintenance argument with a real failure behind it survives review better than one without.

---

pando-genet:
**Correction to my previous comment.** It presented the hash observation as fresh confirmation of this PR's content-not-hash position. It was not fresh — #57 had already established it, from the CLI source, three days earlier:

> the hash's **only** comparison in the entire CLI is `computeSkillFolderHash(skill.path) === existingEntry.computedHash`, used to skip skills already up to date. It compares the *local folder* against the lock, never *fetched upstream content* against a pinned value, and a mismatch causes an install, never an error.

So the four repos sharing hash `82c8da67…` while `ghx`'s content differs does **not** show a guarantee failing. It shows `ghx`'s lock is stale against its own content — the same partial-write state #57 opened with, and #57 named `grilling` as one of the two skills it happened to. I reported a known, already-ruled behaviour as a discovery, which is worse than not reporting it: it invites a re-litigation of a fork this ticket already closed.

What actually holds up, and is worth keeping:

- **A third repo is still carrying that state** days after #57, which is a durability signal about the partial-write case rather than about the hash.
- **The consequence was observed, not just reasoned about.** #57 said "nothing prevents an unreviewed upstream skill change from landing in a downstream repo." Here one had landed, the substituted body lacked its `## Recording` section, and a full planning session produced fourteen rulings and recorded none — silently, because a wrong skill does not announce itself.
- **The substituted text came from `mattpocock/skills`** — untrusted under this PR's allowlist — occupying a slot whose lock entry names `frankify-app/skills`. That is precisely the quarantine lane's case, and it is the part of this instance that speaks to *this* PR rather than to #57.

Still no change requested here; the design already covers it. Correcting the record because a rationale propped up by a misattributed finding is weaker than one without it.

pandoscope/ghx#7 has been reframed as repo-specific remediation.

---

Issue #55, opened 2026-07-25
tools/record.py is not lint-clean under stricter downstream ruff configs


frankify-app:
## Problem

`template/tools/record.py` renders into generated repos, where it is linted by the *downstream* repo's ruff config. On frankify-app/disambiguate (strict config: pydocstyle `D`, bandit `S`, bugbear `B`), a `copier update` adoption of the template immediately fails `prek run --all-files` with **41 errors** in `tools/record.py` (29 unfixable): `D103` missing docstrings on ~20 public functions, `D205` summary-line spacing, `B904` `raise ... from`, `S603`/`S607` subprocess annotations.

So the template violates its own "prek must pass on every commit" rule on arrival in any repo with rules stricter than the template's.

## Options

1. **Lint-harden `record.py` in the template** (docstrings everywhere, `raise ... from`, `# noqa: S603,S607` with justification) so it survives the strictest downstream config in the fleet. Preferred — the file is template-owned, so it should meet template-fleet standards, and downstream diffs stay empty.
2. Template ships a documented convention that template-vendored paths (`tools/`, `scripts/`) get downstream `per-file-ignores` — managed duplication linted under the template's config, not the repo's. (This is the stopgap disambiguate is adopting; see the adoption PR.)
3. Both: harden now, keep the convention for future vendored files.

## Origin

Found while adopting template PR #54 on disambiguate (session flow-test of #53's enforcement — the new per-commit lint rule caught it immediately, which is the rule working as intended).

Issue #52, opened 2026-07-25
Shared glossary terms ship with the scaffold


frankify-app:
## Context

Grilling session 2026-07-25 ruled: canonical definitions of cross-repo vocabulary live in this repo and are distributed to consumers by copier, same mechanism as any other template change — so a definition change rides the same review path as a schema change.

**Membership test (the promotion rule):** a term is promoted into the shared set when it is cross-referenced between repos. When the need arises, the term gets promoted; no static list to maintain.

**Body revised 2026-07-25** after implementation. The original spec called for a `glossary` subtemplate rendered alongside the scaffold. That was built, then withdrawn — implementation exposed costs the original ruling did not have. Superseded design and reasoning below; the original text is in the edit history.

## Design

- **Source:** `template/docs/glossary/*.md`, delivered by the ordinary scaffold render. One copy, one delivery path, no second answers file and no second `copier update`.
- **Marker:** each file carries `<!-- d10e: auto-prune -->`. Copier copies plain `.md` verbatim, so the marker must be in the source — there is no render-time injection.
- **Why the source cannot be pruned:** `disambiguate` walks up and finds the repo-root `docs/glossary/` first, so nothing under `template/` is ever a glossary root. The canonical files are structurally unprunable; every stamped *copy* consents to removal.
- **Why the terms are lintable at all:** they become real terms only once stamped into this repo's own `docs/glossary/`. Hence the self-application gate below.
- **Reachability:** this repo's README links the terms, so they are reachable here. README is repo-owned and **not** part of the render, so a generated repo receives the same terms unlinked — orphaned until it references them or `disambiguate prune` removes them.

## Self-application is now enforced

`tests/test_self_application.py` renders the template and diffs it against the repo root. A stale stamp means the glossary being linted is not the glossary being shipped.

Enforces the **invariant**, not a proxy: any commit order passes, and it catches a stamp commit that a later commit made stale — which a "last commit must be the stamp" rule cannot. Exclusions: copier's `_skip_if_exists` paths are excluded structurally (seeded once, can never match); everything else must match or appear in `DELIBERATE_DIVERGENCE` with a reason, so a new template file forces a decision instead of being silently unadopted.

Measured **7 divergences** on `main` against a documented exception list of **1**. The list now lives in the test and is accurate.

## Tasks

- [x] Shared terms in `template/docs/glossary/`, repo-neutral, link-closed, acyclic
- [x] `auto-prune` marker on every shared term
[...]

PR #66, opened 2026-07-25
feat!: move decision-memory tooling into the guard subtemplate

### LINKED ISSUE #64: record.py `open --use` inherits the attached checkout's current branch, silently mixing sessions
## Problem

`tools/record.py open --use «path»` creates the session branch from **whatever branch the attached checkout happens to be on**, not from the store's default branch. If that checkout still sits on a previous session's branch — the normal state after any session that hasn't been merged yet — every unmerged record from that session rides along into the new session's PR.

Observed today: a session that produced **5 records + 1 proposal** opened a PR containing **29 changed files** — its own 6 plus 16 records belonging to an open PR in the store and 5 from an earlier glossary batch. The PR claimed to be five decisions and was silently proposing twenty-three more, including a `preferences.md` counter change from another session.

The trap is worst in exactly the environment `--use` exists for: managed/remote sessions where cloning is impossible, so the attached checkout is reused across sessions and accumulates branches.

## Why it goes unnoticed

`record.py check` validates the **whole corpus**, not the session's diff — so it reports "N record(s) valid" and the writer concludes the session is clean. Nothing in the tool's output states which files this session actually added.

## Proposed fix

1. **Branch from the store's default branch**, not from `HEAD`: `git fetch origin «default»` then create the session branch off `origin/«default»`.
2. **Or refuse** when the checkout is not on the default branch and no explicit base is given, naming the branch it found — a loud stop beats a silent mix.
3. **Report the session's own diff** at `record` and at session end (`git diff --name-only «base»...HEAD`), so the writer sees exactly what the PR will contain. `check`'s corpus-wide validation is not a substitute.

Fix 3 is worth doing regardless of 1/2 — it is what would have caught this immediately.

## Acceptance

- [ ] `open --use` on a checkout parked on an unrelated branch produces a session branch containing only that session's commits.
- [ ] If the tool cannot determine a safe base, it exits non-zero and names what it found.
- [ ] Session output states the files the session added, distinct from corpus validation.
- [ ] Regression test with a checkout parked on a non-default branch.

## Origin

legacy decision-memory ticket 12, corrected by rebuilding the branch from `origin/main` and cherry-picking the session's own six commits.

### LINKED ISSUE #65: Move decision-memory writer tooling into the guard subtemplate; absorb the store-owned budget/replay layer
## Context

**The subtemplate already exists.** `guard/` landed with #44 (merged 2026-07-22), is selected by `agentic_subtemplate=guard`, and the legacy decision-memory store is already stamped from it (`.copier-answers.agentic.yml`: `agentic_subtemplate: guard`, `_commit: v2.0.0`). This ticket **extends** it — no new subtemplate mechanism is needed.

What is misplaced is the **writer** tooling. `template/tools/record.py` (862 lines) renders into **every** repo generated from the main template, even though it is decision-memory-specific code that an ordinary consumer repo does not need at rest. Meanwhile the store-side budget/replay layer was built as store-owned in a repo that is supposed to be data-only. Both halves should live in the one subtemplate that only ever reaches decision-memory stores.

**This reverses a decision that is currently documented in three places.** Rewriting them is part of the work, not an afterthought:

- `guard/AGENTS.md` — "Data-only — no writer tooling lives here (the recorder lives in the agentic-engineering-template repo)"
- `guard/README.md` — "The writer tool (`record.py`) lives with its consumers in the agentic-engineering-template repo … This repo stays ignorant of which tools write to it."
- the legacy decision-memory store `.github/store/config.py` — "Store-OWNED (not copier-vendored): this file and everything else under `.github/store/` belongs to this repo … The vendored guard next door (`.github/guards/`) stays untouched"

The last one is the design of legacy decision-memory ticket 3, whose branch (`claude/decision-memory-issue-3-hgpbul`, 1 commit / 10 files / 1761 insertions ahead of `main`) is **to be abandoned in that repo** and re-landed here.

### Moves from the main template into `guard/`

| From | To | Note |
| --- | --- | --- |
| `template/tools/record.py` | `guard/tools/record.py` | the recorder; stops rendering into consumer repos |
| `tools/record.py` (repo root) | *deleted* | root self-application; the template repo is not a store |

### Moves from the legacy decision-memory store `claude/decision-memory-issue-3-hgpbul` into `guard/`

| File | Note |
| --- | --- |
| `.github/store/config.py` | repo-local config loader; docstring's ownership rationale is now false |
| `.github/store/budget.py` | token-budget status, `ok`/`warn`/`over` |
| `.github/store/preferences_guard.py` | carve-out label gate + replay-report check |
| `.github/store/replay.py` | masked-replay scoring + two-stream gate |
| `.github/store/tests/test_store.py` | store-layer self-test, run by `preferences-guard.yml` |
| `.github/store/README.md` | |
| `.github/workflows/preferences-budget.yml` | push-to-main measure + pinned compression-due issue |
| `.github/workflows/preferences-guard.yml` | PR gate |
| `.claude/skills/compact-preferences/SKILL.md` | manual compaction skill |
| `store.config.json` | **must get `_skip_if_exists`** — human-adjustable knobs (`budget_tokens`, `replay_window`, label names), same treatment as `preferences.md` |

### Stays in the main template

`DECISION_MEMORY_URL` in `AGENTS.md` / `template/AGENTS.md.jinja`, and the `scripts/doctor.sh` reachability check. These describe how a **consumer** reaches a store; they are not store code. Their wording changes only if the bootstrap answer below changes how a consumer invokes the recorder.

## Open design question — resolve before implementing

`record.py` is **writer-side**: it runs in the *consumer* repo's session and writes into the store. If it only ships to stores, a consumer session needs a way to obtain it, and `open` — the verb that clones the store — is exactly the verb that would no longer be present to run. Candidates:

1. **The store checkout is the tool source.** Session does a plain `git clone "$DECISION_MEMORY_URL" «tmp»`, then runs `python «tmp»/tools/record.py open --use «tmp»`. `--use` already exists for precisely the attached-checkout case (#44), so this is the smallest change — but it demotes `open`'s own clone path to dead code in the common flow.
2. **Bootstrap shim stays in the main template.** A few-line `tools/record` that clones and re-execs. Keeps consumer ergonomics; keeps a (small) decision-memory-specific file in every consumer repo, which is the thing this ticket is removing.
3. **The `grilling` skill carries the bootstrap.** Puts it where the caller already is — but frankify-app/skills `derived/grilling/SKILL.md` does not mention `record.py` at all today and still reads the pre-#44 variable name `DECISION_MEMORY_REPO`, so this option implies a skills-repo change first (frankify-app/skills#11 territory).

Recommend **1** — it needs no new artifact in consumer repos and reuses a flag that already exists — but this is a ruling for a human, and it decides the AGENTS.md wording, so it blocks the rest.

## Removal from already-stamped repos — verified: yes, `copier update` does it

Tested empirically on copier 9.15.2 (this repo's pin is `copier==9.*`) with a synthetic template mirroring the real layout — templated `_subdirectory: "{{ agentic_subtemplate }}"`, two subtemplates, `_answers_file`. v1 stamps `tools/record.py`; v2 `git mv`s it into the guard subtemplate; the project then runs `copier update --trust --defaults --answers-file … --vcs-ref v2`:

- **Unmodified downstream file** → deleted. Post-update tree is `.copier-answers.agentic.yml` + `keep.txt`; `git status` shows ` D tools/record.py`. The emptied `tools/` directory is removed too.
- **Downstream-modified file** (a local edit committed before the update) → **also deleted, silently.** Copier exits 0, no conflict, no `.rej`, and the local edit is gone.

So the answer to "can we remove the distributed copies by stamping?" is yes — with a caveat worth stating in the rollout note: **deletion beats local modification without warning.** Anyone who hand-patched their `tools/record.py` loses that patch with no signal. Relevant because #55 pushed downstream repos toward exactly that kind of local accommodation.

Reach is limited to repos that actually run `copier update` to a version at or past the removal:

| Repo | `_commit` | Has `tools/record.py`? |
| --- | --- | --- |
| frankify-app/disambiguate | v2.0.1 | yes — will be deleted on update |
| frankify-app/meta | v2.0.1 | yes — will be deleted on update |
| frankify-app/skills | f4330ec | no (predates #44) |
| frankify-app/ghx | v1.1.0 | no (predates #44) |
| this repo (root self-application) | — | yes — remove in this PR |

**Blocked by #60** for the actual rollout: `copier update` currently aborts and rolls back in any environment without `gh`, because `_tasks` runs `scripts/doctor.sh` non-advisory. Every consumer adoption of this change runs into that first, unless the operator knows to pass `--skip-tasks`.

### Ruling

- [ ] Resolve the bootstrap question above; record the ruling (it sets the AGENTS.md wording and whether `open`'s clone path survives)

### Template

- [ ] `git mv template/tools/record.py guard/tools/record.py`; delete the root `tools/record.py` self-application
- [ ] Port the ten files from `claude/decision-memory-issue-3-hgpbul` into `guard/`, unchanged in behavior
- [ ] Add `store.config.json` to `_skip_if_exists` under the `agentic_subtemplate == 'guard'` branch, next to `preferences.md`
- [ ] Decide whether `.github/store/` stays a separate directory from `.github/guards/` now that both are vendored. The split's stated rationale (vendored guard vs store-owned layer) no longer holds; keeping two directories is the low-risk option, merging is the honest one. Either way `config.py`'s docstring must be rewritten — the `PREFERENCES_TOKEN_BUDGET` backstop relationship it documents stays true and should be kept
- [ ] Rewrite the three "no writer tooling lives here" / "store-owned" passages in `guard/AGENTS.md`, `guard/README.md`, `.github/store/config.py`
- [ ] Update `guard/README.md`'s writer-flow section for whatever the bootstrap ruling says

### Consumer-facing

- [ ] `AGENTS.md` + `template/AGENTS.md.jinja`: `DECISION_MEMORY_URL` prose currently says "`tools/record.py` (and the `grilling` skill through it)" — the path is no longer local to the consumer repo
- [ ] `scripts/doctor.sh` + `template/scripts/doctor.sh.jinja`: the comment "session-start shallow-clone is the recorder's job" needs to match the bootstrap ruling

### Tests

- [ ] `tests/test_guard_subtemplate.py`: extend `GUARD_FILES` (exact-set assertion per #63) with the moved files; add a byte-identical vendoring check for `record.py` alongside the existing validator check
- [ ] `tests/test_generate_project.py`: drop the `dst_path / "tools" / "record.py"` assertion (line ~470); assert its **absence** in a `template` render instead
- [ ] `tests/test_record_core.py`: repoint at the new location; contract-core coverage unchanged
- [ ] Add a regression test that a `template` render contains no decision-memory writer artifacts
- [ ] Port `.github/store/tests/test_store.py` so it runs in this repo's suite as well as in the store's `preferences-guard.yml`

### Store-side note

- [ ] `.claude/skills/compact-preferences/` renders into the store as a **plain directory**, not the `.claude/skills → ../.agents/skills` symlink the main template uses, and the guard subtemplate runs no skills-install task (`_tasks` are gated on `agentic_subtemplate == 'template'`). Confirm this is intended rather than an accident of the port

### Rollout

- [ ] Close legacy decision-memory ticket 3 and its branch as superseded, pointing at this ticket. The legacy decision-memory store's next `copier update` re-acquires the same files as vendored — verify the tree is equivalent, with `store.config.json` preserved by `_skip_if_exists`
- [ ] Adoption PRs for disambiguate and meta, each carrying a note that `tools/record.py` is removed by the template and that local edits to it are dropped without warning
- [ ] Land or work around #60 first, or document `--skip-tasks` in the adoption PRs

## Acceptance

- [ ] A `template` render contains no `tools/record.py` and no other decision-memory writer artifact
- [ ] A `guard` render contains the recorder, the guard, the budget/replay layer, the compaction skill and `store.config.json`, and the exact-set assertion in `test_guard_subtemplate.py` covers all of them
- [ ] Re-rendering a `guard` project that has an edited `store.config.json` and `preferences.md` preserves both
- [ ] `copier update` on a repo currently holding `tools/record.py` removes it and leaves no `.rej`
- [ ] The store's `preferences-budget.yml` and `preferences-guard.yml` behave identically to the issue-3 branch after being vendored
- [ ] No file in the repo still claims the store holds no writer tooling
- [ ] Recording end-to-end works from a consumer repo through whichever bootstrap path is ruled

## Non-goals

- No behavior change to the recorder, the guards, the budget maths, or the replay scoring — this is a relocation. Behavior bugs stay on their own tickets (#64, #55)
- No change to the `grilling` skill in frankify-app/skills unless bootstrap option 3 is chosen
- No change to `preferences.md`'s seeded-once treatment
- Not the glossary subtemplate (#52) — separate distribution, separate ticket

## Considered alternatives

| Option | Rejection reason |
| --- | --- |
| Leave `record.py` in the main template, only move the issue-3 layer | Leaves the actual complaint unaddressed: every consumer repo still carries 862 lines of store-specific code it never runs, and #55 (not lint-clean under downstream ruff configs) stays live |
| Keep the issue-3 layer store-owned as designed, move only `record.py` | Two decision-memory stores would then drift on budget and replay logic exactly as the docs argue schema must not drift; the reason for vendoring the guard applies verbatim to the layer above it |
| Give `record.py` its own installable package instead of a subtemplate | Larger change, new release surface, and the fleet has exactly one store today. Revisit if a second writer implementation appears |
| Delete the distributed copies by hand in each consumer repo | Copier removes them mechanically on the next update (verified above), and hand-deletion leaves the answers file claiming a template version whose render no longer matches |
| Third subtemplate for the writer, separate from `guard` | Only meaningful if a repo wants the recorder without being a store. No such repo exists, and the whole point is that the recorder ships to stores |

## Origin

Session 2026-07-25. Code inventory across this repo, the legacy decision-memory store (`claude/decision-memory-issue-3-hgpbul`), and the four template consumers; copier deletion behavior established by direct experiment rather than from documentation.

frankify-app:
Closes #65. Closes #64.

Reverses the distribution direction for decision-memory code: writer tooling and the preference-set lifecycle stop being spread across every repo generated from the main template, and are stamped only into decision-memory stores through the subtemplate — which is renamed `guard` → `decision-memory` in the process, since it stopped being only a guard.

## What moves

**Out of the main template.** `template/tools/record.py` becomes `decision-memory/tools/record.py`; the root self-application is deleted, since the template repo is not a store. 862 lines of store-specific code stop rendering into every consumer repo, where they were never run and were linted by that repo's config (#55).

**Out of the legacy decision-memory store.** The budget reporter, the carve-out/replay PR guard, their two workflows, the compaction skill, the seeded config and the layer's self-test move from that repo's now-closed ticket 3 into the subtemplate. They were built store-owned; one store can own them, N stores cannot — each would drift from the schema the vendored guard enforces, which is the argument that made the record guard vendored to begin with.

`.github/store/` stays a separate directory from `.github/guards/`. Both are vendored now, but they answer to different things: `guards/` gates the record corpus, `store/` runs the preference-set lifecycle on top of it. Every file that claimed "store-owned, not copier-vendored" now states its source and update mechanism, per the managed-duplication rule in AGENTS.md.

`store.config.json` is the one exception — seeded once via `_skip_if_exists`, so a store tunes `budget_tokens` and `replay_window` without `copier update` reverting the choice.

## Subtemplate rename

`agentic_subtemplate=guard` becomes `agentic_subtemplate=decision-memory`. The name described one of the five things it ships. `decision-memory` is the glossary term for what it stamps (#52 shipped `docs/glossary/decision-memory.md`), and the answer reads `agentic_subtemplate: decision-memory` — where `decision-memory-template` would stutter against the key it answers. The prose is unified too: the docs previously said "the guard subtemplate" in some places and "the store subtemplate" in others.

The CI guard keeps its own name — `.github/guards/`, `guards.py` and `guards.yml` are untouched, because that layer really is a guard.

**This breaks the already-stamped store.** The legacy decision-memory store's answers file carries `agentic_subtemplate: guard`, and `_subdirectory` resolves to `{{ agentic_subtemplate }}`, so its next update needs the answer migrated first. Tracked in that repo's ticket 16.

## Recorder changes

Three edits to a file already being moved, ruled in the grilling session recorded on #65:

- **The recorder operates on the checkout it lives in.** `store_root()` resolves it; nothing is passed in, searched for, or configured. `open` no longer clones — the session clones first and runs the copy that arrives with it.
- **The session branch is based on `origin/«default»`, not `HEAD`** — closes #64. Verified against the live checkout in this session: the old behavior would have based a session on `67bca1a` rather than `origin/main` at `2598ad9`, putting 10 files and 1761 insertions of unrelated work into the PR.
- **Each record is pushed as it lands.** Sessions run in ephemeral clones, so holding records until `submit` made the clone's survival load-bearing. A failed push exits non-zero rather than leaving a record silently unpublished.

## Removal from already-stamped repos

Verified empirically on copier 9.15.2 — first against a synthetic template mirroring this layout, then end to end against a throwaway copy of the real disambiguate repo: `copier update` **deletes** a file removed from the template, removes the emptied `tools/` directory, and leaves no `.rej`. It also deletes a downstream-**modified** copy silently, exit 0, no conflict — a hazard the adoption tickets call out.

Consumer cleanup is deliberately **not** in this PR: the consumers are pinned to released tags, so removing their copies needs a `copier update` against a ref that contains the removal. Tracked per repo — frankify-app/disambiguate#53, frankify-app/meta#8, frankify-app/skills#21, frankify-app/ghx#4 — each noting that `v2.1.0` still ships the recorder, so `skills` and `ghx` must not stop there or they would acquire the file for one release before losing it.

## Obstacles

- **#60 blocks adoption everywhere, including CI here.** `tests/test_e2e.py::test_e2e_prek_install_registers_git_hooks` and `::test_e2e_smoke_full_render_runs_tasks` fail locally with `Command 'bash scripts/doctor.sh' returned non-zero exit status 1` — doctor exits 1 because `gh` is absent, and a non-zero `_tasks` entry aborts the whole copier run. Confirmed pre-existing by stashing this branch's changes and reproducing at `main`; both pass in CI, which has `gh`. 110 other tests pass locally.
- The store layer's self-test asserts a real `decisions/` corpus. The same file now runs both in a store and here, so that one check skips where no corpus exists; the fixture tests still cover the harness.
- `ruff` reformatted the ported files on arrival — they are template-owned now and linted by this repo's config, so the vendored copies are no longer byte-identical to their decision-memory originals.
- The session fixture in `tests/test_record_session.py` needs the store's own `.gitignore`, because importing the recorder in-process writes `__pycache__` and `open` refuses a dirty worktree.

## Deviation from the ticket as filed

#65's Non-goals said "no behavior change to the recorder — this is a relocation". Q4 of the grilling session overrode that deliberately: the three edits are small, the file was already moving, and splitting would mean writing the bootstrap recipe into AGENTS.md twice. Recorded in full on #65.

## Review round

- Removed-feature mentions are erased rather than annotated, and the general rule is now in `template/AGENTS.md.jinja` § Rules.
- The AGENTS.md entry states what is, not what stopped being; the entry is 4 lines shorter than the first draft.
- Root `AGENTS.md` and `scripts/doctor.sh` are no longer hand-edited — the template-only change and the `chore: reapply template to self` stamp are separate commits, per `docs/conventions.md` § Template-First Changes.
- The compaction skill moved to `.agents/skills/` with `.claude/skills` as the symlink bridge, matching the main template.

## DECISION markers

- `DECISION:` (`record.py`, `store_root`) — the recorder always operates on its own checkout; it is stamped into stores by the subtemplate, so "which checkout?" has exactly one answer and needs no flag, env var or search to resolve.
- `DECISION:` (`record.py`, `check_store_checkout`) — origin is matched by owner/repo tail, not textually, because managed environments rewrite remotes through a local proxy and a rewritten origin never equals the configured URL.
- `DECISION:` (`record.py`, `cmd_open`) — the session branch is based on `origin/«default»`, never on `HEAD`; a checkout parked on a previous session's unmerged branch is the normal state of any reused checkout.
- `DECISION:` (`record.py`, `push_session`) — every record is pushed the moment it is committed, so the clone holds nothing the remote does not and its location stops being a durability question.

https://claude.ai/code/session_013bmY2VqQ4S6gR8NJ7TjW1G


guard/tools/record.py
```diff
@@ -29,8 +28,20 @@
   propose  write a preference-rule proposal file with its commit
 
 Configuration: DECISION_MEMORY_URL (full git URL of the data repo;
-never commit it anywhere public). Verbs after `open` find the clone
-via DECISION_MEMORY_DIR or --dir.
+never commit it anywhere public).
+
+The recorder ships inside the store, so it always operates on the
+checkout it lives in — no --dir, no --use, nothing to point at
```
frankify-app: This commit is removing `--dir` and `--use`, right? Then we can just remove them completely, no one in the future would care that they once existed. Please record this as a feedback for AGENTS.md as well, something like: "when removing a feature, ensure its mention gets erased from all other places too"

template/AGENTS.md.jinja
```diff
@@ -123,7 +123,16 @@ Code-specific skills:
...
+- `DECISION_MEMORY_URL` — FULL git URL of the decision-memory repo the `grilling` skill records decisions to. A full URL rather than an owner/repo slug, so the hosting stays swappable. Recording requires this env var in the agent's execution environment; the recorder and the skill read exactly this name (shared contract — renaming either side breaks recording silently). Never hardcode, commit, or echo the value into artifacts. Unset → grilling still works, skips recording, and says so. Where to set it: local sessions → shell profile / user-level agent settings; remote or cloud sessions → the environment's configuration; CI → a repository secret. `scripts/doctor.sh` warns when it's unset and checks reachability when set.
+
+  The recorder is NOT in this repo — it is store tooling, stamped into decision-memory repos only, and it operates on the checkout it lives in. To record, clone the store and run the copy that arrives with it:
```
frankify-app: Again, we don't need to say what is not the case (it used to be but that doesn't matter anymore) we just say what is now. So: "Use the recorder present in the decision-memory repo" (or something like that). Remember, token budget in AGENTS.md is very scarce

AGENTS.md
```diff
@@ -115,7 +115,16 @@ Code-specific skills:
...
+
+  The recorder is NOT in this repo — it is store tooling, stamped into decision-memory repos only, and it operates on the checkout it lives in. To record, clone the store and run the copy that arrives with it:
+
+  ```bash
+  git clone "$DECISION_MEMORY_URL" <dir>
+  python <dir>/tools/record.py open      # behavior doc: record.py --help
+  ```
+
+  Clone fresh per session rather than reusing an attached checkout: a fresh clone is clean and on the store's default branch, which is what keeps a session's PR to that session's own records. Each record is pushed as it lands, so the clone is disposable.
```
frankify-app: This file should not be changed by this commit, but rather by stamping the template in the last commit of the PR

guard/tools/record.py
frankify-app: Yes — `--dir`, `--use` and `DECISION_MEMORY_DIR` are all removed. Erased the mentions: the docstring now just states the invariant, and the `store_root` DECISION marker says "needs no flag, env var or search to resolve" instead of naming the three things it replaced. One more was hiding in `tests/test_record_core.py`'s docstring; gone too. `grep` for either flag across the repo now returns nothing.

The migration detail stays in the commit's `BREAKING CHANGE:` footer, which is where someone upgrading actually looks.

Rule recorded in `template/AGENTS.md.jinja` under `## Rules`:

> When removing a feature, erase every mention of it — docs, help text, comments, tests. Don't leave "no longer supported" notes: readers who never knew it existed pay to learn it did. State what is, not what stopped being. (Migration notes belong in the commit's `BREAKING CHANGE:` footer, which is where someone upgrading looks.)

---

AGENTS.md
frankify-app: Right — I edited the render output directly, which `docs/conventions.md` § Template-First Changes explicitly forbids. Restructured to follow the documented route: the preceding commit is template-only, and `chore: reapply template to self` adopts the rendered `AGENTS.md` and `scripts/doctor.sh` verbatim from a `copier copy --vcs-ref HEAD` render.

`tests/test_self_application.py` passes, so root and render agree.

---

template/AGENTS.md.jinja
frankify-app: Reworded to your phrasing, positive only:

> To record, use the recorder in the decision-memory repo — clone it fresh per session and run the copy that arrives with it, which operates on its own checkout:

Dropped "The recorder is NOT in this repo — it is store tooling, stamped into decision-memory repos only" entirely, and folded the clone-fresh rationale into one sentence instead of a paragraph. Net for the entry is 4 lines shorter than what you reviewed.

Kept the two-line code block: it is the whole instruction, and prose describing a `git clone` costs more tokens than the command.

---

PR #80, opened 2026-07-26
feat(decision-memory): preference extraction, closed store-layer seams, calibrated replay gate

### LINKED ISSUE #78: decision-memory: preference extraction — records to candidate rules
Moved here from the store repo, where it was originally filed. Extraction is enforcement-adjacent tooling, and every enforcement surface in a store is copier-vendored from this subtemplate — the store owns only `preferences.md`, `decisions/`, and its own config. A store-side extraction skill would be a parallel store-owned layer with no update path, exactly the seam that pulled the budget/compaction work here.

## Context

Grilling sessions append decision records to a store. Rules in `preferences.md` are the injected predictive layer. Extraction is the bridge, and it is currently missing: the corpus grows, the rule set does not, and every record is recorded `prediction_stream: cold` because no rule was ever there to drive it.

**Extraction precedes compaction.** There is nothing to compact until rules are extracted from records, and the replay gate (#79) cannot measure a rule set that no record was ever scored against. That ordering is the reason both land together.

## Design

- Extraction runs manually via a skill, like compaction — industrialize when it is boring.
- Reads: the records in scope + the whole corpus as evidence + current `preferences.md` + `proposals/`. Never `decisions/` for writing — extraction is read-only over history, always.
- Per candidate pattern, **exactly one** of:
  - matches an existing rule → counter increment + last-confirmed date (`pref-confirm:`)
  - contradicts an existing rule → drift flag proposing conditionalize-or-retire (`pref-drift:`); never a silent overwrite
  - genuinely new → propose a rule in conditional, falsifiable form (`pref-proposal:`)
- **Correction events first.** A `"N, but actually because…"` ruling (`correction: true`) carries the decider's true reason where the model's guess was wrong — the highest-signal record kind in the corpus.
- **Prediction misses are the second queue.** Each miss must refine an existing rule, split one, or spawn a candidate. A miss that does none of those is logged as unexplained — unexplained is a state, not a silence.
- A rule only ever confirmed by choices its own recommendation caused has zero independent evidence. Extraction flags such rules and never strengthens them; provenance is already in the records (`rules_cited` + `chosen_slot`).
- Output: one commit per operation. Human merge is promotion — merging a proposal file is not.

## The watermark is a commit

Revised twice from the design as originally filed, which specified a batch pass over everything since a `last-extracted` marker file. **There is no marker file.** An intermediate version used the PR diff as the boundary; that is also gone.

Every pass ends with a **`pref-extract:` commit**, so the scope is everything recorded since the last one — derived from history rather than tracked beside it. Nothing to advance, nothing to keep in sync, nothing to conflict when two branches run a pass at once, and the message carries what the pass found.

Two properties fall out of deriving rather than storing:

- **It self-heals.** A session merged without a pass is not lost; the next pass simply reaches further back. A stored watermark that was never advanced looks identical to one with nothing to do. This is why the PR-diff boundary was not enough on its own: correct, but a single bypass lost those records permanently.
- **Bootstrap is not a special case.** No `pref-extract:` commit anywhere means the watermark is the beginning of the corpus, so the first pass covers everything by construction. There is no `--all` mode because there is nothing for it to do.

### An empty pass commits too

This is what makes detection reliable rather than merely convenient. A pass that finds nothing produces no proposal and no counter bump, so a watermark keyed on outcome commits would stall every time extraction legitimately had nothing to say. `pref-confirm:` is worse than useless for the purpose: `record.py submit` emits those, so it would move the watermark for a pass that never ran — and a false watermark skips records rather than re-reading them. Hence a dedicated type nothing else produces, and `--allow-empty`.

### Scope is not evidence

Running the pass per session does not make it per-session in the sense the original design was protecting against:

- **Scope** — what the pass must act on — is the records since the watermark.
- **Evidence** — what the pass may reason from — is the whole corpus, shipped in the batch as `history`.

Cross-session repetition, the evidence a single session cannot see, stays fully available. A pattern appearing once in the scope and twice in `history` is a three-record pattern.

### The gate turns "recoverable" into "caught"

A missed pass being recoverable is not the same as it being recovered — nothing would prompt the recovery, which is exactly how a 17-record backlog accumulates. So a PR adding records must contain a `pref-extract:` commit with no record added after it (extraction is the last step of the pass). The check is purely positional: the commit's PLACE in history is its claim, so there is nothing to enumerate, nothing to keep in sync with the diff, and nothing copyable from another branch.

### Known constraint

**The repo must not squash-merge.** Squashing rewrites subjects and the watermark is a subject in history. Merge commits and rebase both preserve it.

## Tasks

- [ ] `.github/store/extraction.py` — `status` and `batch` over the derived watermark, plus the positional pass check. Stdlib only.
- [ ] `.agents/skills/extract-preferences/SKILL.md` — the procedure driving it.
- [ ] `pref-drift:` and `pref-extract:` commit types in the commit lint and `docs/conventions.md`.
- [ ] Pass enforcement in the PR guard: a PR adding records must close with a pass commit.
- [ ] Tests against a fixture store with real git history, in template CI.
- [ ] One-off bootstrap pass over the corpus that predates extraction.

## Open questions

- Hit-only batches: extract confirmations, or only process misses and corrections? Confirmations are cheap counter bumps, so probably yes — and a pure-hit batch is the obvious first candidate for auto-merge someday.

## Constraints

- **This template repo is public.** Never write a store repo's URL, name, owner, issue numbers, or record slugs into any file here. Consumers reference their store through `DECISION_MEMORY_URL` only.
- `decisions/` append-only is absolute. Extraction reads; it never writes there.
- Stdlib only, no dependencies — the vendored copy must keep working if this repo disappears.

## Non-goals

- No scheduled extraction — manual trigger, human-merged.
- No embedding/RAG retrieval.
- No change to the grilling skill.

### LINKED ISSUE #79: decision-memory: close the seams the store-layer move opened, and calibrate the replay gate
Follow-on to #66 / #74, which moved the preference budget + compaction tooling out of a store repo and into this subtemplate. The move itself has landed: `.github/store/{config,budget,preferences_guard,replay}.py`, the two workflows, the compaction skill and `store.config.json` are all vendored, and a store has been restamped from `v3.1.0`.

The move was the prerequisite, not the point. It was undertaken because three seams could only close on this side, and none of them has closed yet.

## The seams

1. **Compaction commits are typed `pref-promote:`** because that is the only vendored type permitted to remove lines from `preferences.md`. Compaction is not promotion. The human-only property of promotion is currently preserved by the merge gate rather than by commit authorship, which works and reads wrong. A dedicated `pref-compact:` type is the fix, and it is a change to the commit lint and to `docs/conventions.md` — both vendored, both only editable here.
2. **Nothing in the vendored `README.md` / `AGENTS.md` / `docs/` points at the tooling.** That was true store-side because editing those files there fights `copier update`. It is still true here because nobody has written the pointer yet. A human reading a store's `README.md` today cannot discover that a budget, a carve-out gate, a replay harness and a compaction skill exist in the repo they are looking at.
3. **The token budget is checked twice.** `decision_validator.check_preferences_budget` enforces a hard 2000-token backstop; `.github/store/budget.py` separately checks the repo-local `budget_tokens`, which `config.py` currently forbids from exceeding the backstop — because the vendored check would fail the PR first. That restriction is a workaround for having two authorities. One vendored check reading the repo-local value removes both the second authority and the workaround.

## The gate is not yet trustworthy

The compaction harness was run end-to-end against a real store's corpus (17 records, `preferences.md` at ~159 tokens, 8% of budget). Step 5 correctly found nothing to compact — three rules, no overlap, no supersession. So it was run instead as a **null test**: two independent subagent runs over the *same* rule set, blind, neither permitted to read `decisions/`.

| | run A | run B |
| --- | --- | --- |
| preference-driven | n=3, 3 hits, **1.00** | n=5, 5 hits, **1.00** |
| cold (control) | n=14, 13 hits, 0.93 | n=12, 11 hits, 0.92 |
| slot predictions | — | identical to A on all 17 |

`gate(A -> B)` returned **pass** — by luck, not by design:

1. **The gated denominator is unstable.** Both runs picked the same slot on all 17 cases but disagreed on whether a rule drove it for 4 of them, so n moved 3 -> 5 under a null change. At that size one case flipping swings the hit rate 20-33 points.
2. **The corpus has no preference-driven history.** All 17 records are recorded `prediction_stream: cold`, so every rule citation registers as a stream shift and the gated stream is built entirely out of the candidate's own labelling. That is the extraction gap (#78), visible in the data.
3. **The absolute number is inflated.** `chosen_slot` is 1 in 14 of 17 records, so a blind "always slot 1" scores 82%; both runs scored 16/17 = 94%. Masking strips `role` and `rules_cited` but leaves slot ordering, and slot 1 is the prediction slot by convention. The leak is symmetric across baseline and candidate, so before/after comparison stays valid — the headline hit rate does not.

Both runs missed the same case, and missed it the same way: they predicted the prediction slot, while the decider had answered in free text.

### Close the seams

- [ ] Add a `pref-compact:` commit type to the commit lint and to `docs/conventions.md`, permitted to remove `preferences.md` lines. Retire the `pref-promote:` overload in the compaction skill.
- [ ] Point `README.md` and `AGENTS.md` at the budget + compaction (+ extraction) flow, in the subtemplate.
- [ ] Make the vendored budget check read the repo-local `budget_tokens` so the "may not exceed the vendored backstop" workaround can go. ONE token-count authority (`estimate_tokens`), ONE budget check.

### Calibrate the gate

- [ ] **Small-n verdict.** A `min_gated_cases` knob; below it, `gate` returns `insufficient-evidence` rather than `pass`, and CI accepts that only with an explicit waiver label — so a human owns an unvalidated compaction instead of a green check implying validation. On a corpus with no extracted rules yet, every compaction needs the waiver. That is the honest state, and it should be visible.
- [ ] **Slot-ordering leak.** Deterministically permute each case's slots (seeded by record id) when building cases, and un-map in scoring, so the number measures rules rather than ordering. The permutation must be derivable rather than shipped — a mapping in the cases file would tell the predicting agent which slot is slot 1.
- [ ] Re-run the null test once extraction (#78) lands and records start carrying `prediction_stream: preference-driven`. The gate is trustworthy when a null change no longer moves the gated n.

### Verify

- [ ] Tests run in template CI against a **fixture store** (throwaway `decisions/` + `preferences.md`), so the guard and harness are testable without any real principal's private store. Today the real-corpus checks silently skip here.
- [ ] `copier update` into a store repo produces the vendored files cleanly, reviewed there as a normal PR diff.

## Layout decision (already taken, recorded here)

`.github/store/` stays a second vendored directory rather than folding into `.github/guards/`. The two have different audiences: `guards/` is the record contract shared with the writer tool, `store/` is the preference-set lifecycle. The unification the ticket actually demanded is of *authorities*, not of directories — one `estimate_tokens`, one budget check — and that is the third task above.

`store.config.json` stays store-owned: the budget is per-principal and human-adjustable. The subtemplate ships the default; `copier update` must never clobber a store's chosen value (`_skip_if_exists`).

## Constraints

- **This template repo is public.** Never write a store repo's URL, name, owner, issue numbers, or record slugs into any file here. Consumers reference their store through `DECISION_MEMORY_URL` only.
- `decisions/` append-only is absolute and gets NO carve-out. The carve-out label applies to `preferences.md` lines only.
- Stdlib only, no dependencies — the vendored copy must keep working if this repo disappears.

## Depends on

- #78 (extraction). The replay gate cannot be calibrated against real numbers until records carry a preference-driven stream. Ship both together; calibrate once real extraction output exists.

## Non-goals

- No scheduled compaction — manual trigger, human-merged.
- No embedding/RAG retrieval tooling.
- No change to the grilling skill.

## References

Deliberately unlinked: the originating ticket and the store-side reference implementation both live in a private store repo, which is never named here. Anyone with access reaches them through `DECISION_MEMORY_URL`.

### LINKED ISSUE #81: decision-memory: ingestion gate — dedup, re-decision links, false-cold, ref completeness
A gate run over drafts + store **before** ingestion, in the store subtemplate's tooling, plus the skill that turns its output into a decision.

Everything here shares one trigger and one reason for urgency: **drafts are mutable and records are not.** `decisions/` is append-only with no carve-out, so anything not fixed at ingestion is frozen permanently. That is what separates this from the rest of the analysis roadmap (#82), whose tickets are gated on record volume and can wait.

## 1. Similarity gate: dedup + re-decision link wiring

Concrete driver: two import issues in a store hold drafts extracted from the *same* session by different runs (branching), so the same underlying decision appears twice with different wording. Ingesting both would mint two immutable records for one ruling.

- Cluster records by question/option token similarity, with an `artifact_ref` corroboration boost.
- Classify similar pairs mechanically:
  - same `preference_set.commit` + same `chosen` → **duplicate** (candidate for discard to `discarded-drafts.json`, never deleted)
  - distinct provenance, no `related`/`supersedes` link → **re-decision** (candidate for adding the link to the draft — records are immutable, so ingestion is the only moment an edge can be written)
  - partial provenance match → **uncertain** (human adjudication)
  - already linked → informational
- Output: worst-first clusters with field-level diffs; `--json` mode.

## 2. False-cold provenance check

`docs/conventions.md` already states that `preference_set.commit` content-addresses the active set so "false cold claims are detectable provenance defects" — and nothing has ever run the check.

- For each draft with `prediction_stream: cold`, load `preferences.md` at the pinned commit and flag rules whose text plausibly matches the question/options (token overlap; advisory, human confirms).
- A confirmed false-cold is fixed in the draft (restream + `rules_cited`) before ingestion.

## 3. `artifact_ref` completeness checklist

Emit a completeness tier per draft (complete / partial / null) so enrichment happens in the same adjudication session as dedup. No auto-guessing of SHAs — conventions forbid it.

## 4. Containment as a second surfacing channel

Added after real data showed the gap. Symmetric similarity is structurally blind to **one draft re-extracted as two**: jaccard divides by the union, so a bundle scores low against each half even when the half sits entirely inside it.

Measured on a real split: **jaccard 0.34** against the half that shares vocabulary, **containment 0.68**. Cost: exactly **1 flagged pair in 136** on a real 17-record corpus, and that pair was a true relation.

Honest about the limit: pairwise comparison cannot *assert* a split, and the second half of that same bundle shares so little vocabulary (0.17) that nothing pairwise finds it. Surfacing one half brings a human to the cluster, which is the actual goal. The report names which channel surfaced a pair, so a containment-only hit reads as the split signature it is.

## 5. `/adjudicate-drafts` — the bridge from clusters to a decision

The gate finds clusters and deliberately does not resolve them: every resolution trades one loss against another, and picking the trade is the human's job. The missing piece was making that call cheap.

Per cluster the skill states the resolution, what that way costs, and what becomes irreversible at ingestion — **with a recommendation**, because a list of options with no position is work handed back rather than a decision made easy. Five resolutions, each with its cost: discard one, keep both linked, keep both unlinked, merge, or treat a split as real.

It also names why this is not cosmetic. **An undetected duplicate double-counts as evidence.** Extraction reads cross-record repetition as the signal that a pattern is a principle rather than a one-off, so one ruling recorded twice manufactures the exact evidence the preference set is supposed to earn — and the replay harness then scores the same case twice.

Rules it inherits: read both drafts before proposing, because the side carrying the only **stated** operative reason is the side whose discard destroys evidence; never delete, discard to `discarded-drafts.json` with a reason; never resolve silently, since "genuinely independent" is a finding rather than a non-answer; re-run the gate afterwards, because a cluster still reading `duplicate` is the one thing ingestion cannot undo.

## What real draft data changed

Validated against two real batches — the same session extracted twice. Two defects were invisible to synthetic tests:

- **Drafts are keyed by `slug`, not `id`.** The recorder mints IDs at ingestion, which is *after* this gate runs by definition, so a gate that only understood `id` printed `None` for every draft it exists to check. Drafts also carry batch-local `related_slugs` / `supersedes_slug`, so an edge already written in a draft read as *missing*.
- **Answers were compared for exact equality**, which called every reworded duplicate "a different answer" and dropped all six real pairs into `uncertain`. Fixed by comparing the **chosen option's label** (identical ruling: label 1.00, prose 0.30) fuzzily — and never the slot number, since two runs listing the same options in a different order gave one ruling slot 1 in one batch and slot 3 in the other. Strictly like with like: label↔label, prose↔prose, never across.

Result on the real batches: **5 duplicates, 1 uncertain, 1 split surfaced by containment, 0 false positives**, plus 39 of 41 drafts flagged for a null `artifact_ref`.

## Constraints

- **Read-only.** The gate writes nothing to the store or the drafts; it reports, the skill proposes, a human acts.
- Stdlib only, no dependencies — the vendored copy must keep working if this repo disappears.
- **This template repo is public.** Never write a store repo's URL, name, owner, issue numbers, or record slugs into any file here.

## Acceptance

- [x] Running the gate on a drafts file + store checkout produces the ranked cluster report, with no writes to either.
- [x] Output includes a `false-cold?` section listing draft id, matched rule line, and overlap evidence.
- [x] Output lists each draft's `artifact_ref` completeness tier.
- [x] Splits surface via containment, with the channel named in the report.
- [x] `/adjudicate-drafts` proposes a resolution per cluster with costs and a recommendation.

## Later, deliberately not now

An advisory version of the same similarity module in the vendored CI guard — warn (never block) when a new record is highly similar to an existing one with no link — is on the roadmap in #82, gated on this heuristic having a measured false-positive rate against a real batch. That measurement now exists (0 false positives on 41 drafts; 1 true-relation hit in 136 store pairs), so the trigger is close.

frankify-app:
Closes #78. Closes #79. Closes #81.

Three tickets, one PR, because they share one vendored directory and one blocked draft batch. #79 depends on #78 — the replay gate cannot be calibrated against a corpus whose every record is `prediction_stream: cold`, which is what a store without extraction has. #81 is upstream of both: it gates the drafts that would become that corpus.

## What was already done

The move itself landed in #66 / #74 — `.github/store/`, the two workflows, the compaction skill and `store.config.json` are vendored, and a store has been restamped from `v3.1.0`. This PR is the part the move existed to enable.

---

## Ingestion gate (#81)

`.github/store/similarity.py`, read-only, run over a drafts file plus the store **before** ingestion. Everything it checks shares one deadline: **drafts are mutable and records are not.** `decisions/` is append-only with no carve-out, so whatever is not fixed at ingestion is frozen permanently. That is what separates this from the analysis roadmap (#82), whose passes read immutable history and can wait for volume.

- **Dedup** — two extraction runs over one session produce the same ruling twice, worded differently; ingesting both mints two immutable records for one decision.
- **Re-decision links** — a `related`/`supersedes` edge cannot be added after ingestion without violating append-only, so ingestion is the only moment an edge can be written.
- **False cold** — `docs/conventions.md` has always said `preference_set.commit` content-addresses the active set so a false cold claim is a detectable provenance defect. Nothing ran the check. It runs now, against the *pinned* set, advisory.
- **`artifact_ref` completeness** — chat drafts carry null/partial refs by design and are enriched once commits exist; SHAs are never guessed.

Classification is by **provenance**: same `preference_set.commit` + same answer is a duplicate, distinct commits are a re-decision, an existing link outranks both. Chat drafts carry null provenance by design, so absence is never read as "distinct" — it falls back to the session key, then the answer, and lands in `uncertain` rather than guessing.

### What real data changed

Validated against two real draft batches — the same session extracted twice by different runs, the exact case this exists for. Two defects were invisible without it:

**Drafts are keyed by `slug`, not `id`.** The recorder mints IDs at ingestion, which is *after* this gate runs by definition, so a gate that only understood `id` printed `None` for every draft it exists to check. Drafts also carry batch-local `related_slugs` / `supersedes_slug` rather than resolved links, so an edge already written in a draft read as *missing*.

**Answers were compared for exact equality**, which called every reworded duplicate "a different answer" and dropped all six real pairs into `uncertain`. Fixed by comparing the **chosen option's label** (on an identical ruling: label 1.00, prose 0.30) fuzzily at 0.5 — and never the slot number, since two runs listing the same options in a different order gave one ruling slot 1 in one batch and slot 3 in the other. Strictly like with like: label↔label, prose↔prose, never across, because one record's option label routinely echoes the other's question.

Result on the real batches: **5 duplicates, 1 uncertain, 0 false positives**, plus 39 of 41 drafts flagged for null `artifact_ref`. The uncertain pair is a genuine judgement call. A 1→2 split (one draft re-extracted as two) scores 0.34 and stays below threshold — pairwise similarity structurally cannot see a split, a known limit rather than a bug.

---

## Extraction (#78)

`decisions/` records what happened; `preferences.md` tells the next session what to expect. Nothing connected them.

- **`.github/store/extraction.py`** — `status` and `batch` over the derived watermark, plus the positional pass check the guard calls. The read side; the skill does the judgement.
- **`.agents/skills/extract-preferences/SKILL.md`** — per candidate pattern, exactly one of: bump a counter, flag drift (`pref-drift:`, a proposal file — never a silent overwrite), or propose a rule.
- **Four queues in descending evidence order** — `corrections`, `misses`, `refinements`, `confirmations`. Every miss must refine a rule, split one, or spawn a candidate; a miss that does none is logged as unexplained, because unexplained is a state and not a silence.
- **Rule-driven acceptances are flagged and never counted** — a rule that cited itself into the prediction slot and got that slot chosen caused the choice it would be credited with predicting.
- Extraction reads `decisions/` and never writes there.

### The watermark is a commit

Revised twice from the ticket as filed: a `last-extracted` marker **file**, then the **PR diff**, now git itself. Both earlier designs are gone, and the commits are in this branch's history rather than squashed away.

Every pass ends with a `pref-extract:` commit, so the scope is everything recorded since the last one — derived from history rather than tracked beside it.

- **It self-heals.** A session merged without a pass is not lost; the next pass reaches further back. That is what the PR-diff boundary could not do — correct, but one bypass lost those records permanently.
- **Bootstrap is not a special case.** No `pref-extract:` commit anywhere means the watermark is the beginning of the corpus. There is no `--all` mode because there is nothing for it to do.
- **An empty pass commits too** (`--allow-empty`), which is what makes detection reliable. A pass that finds nothing produces no proposal and no counter bump, so a watermark keyed on outcome commits would stall every time extraction legitimately had nothing to say. `pref-confirm:` is worse than useless — `submit` emits those, so it would move the watermark for a pass that never ran, and a false watermark *skips* records rather than re-reading them.

**Scope is not evidence.** The pass ACTS on the records since the watermark and REASONS from the whole corpus, shipped as `history`. Cross-session repetition stays fully available.

**The gate turns "recoverable" into "caught."** Nothing would prompt a recovery, which is how a backlog accumulates. So a PR adding records must contain a `pref-extract:` commit with no record added after it.

**Known constraint:** the repo must not squash-merge. Squashing rewrites subjects and the watermark is a subject.

---

## Seams closed (#79)

**One budget authority.** `PREFERENCES_TOKEN_BUDGET` was a hard backstop and `config.py` then forbade `budget_tokens` from exceeding it — two authorities plus a workaround for having two. The record guard now loads `store.config.json` and enforces whatever the store chose; the constant is the default, not a ceiling.

**`pref-compact:`, `pref-drift:`, `pref-extract:`.** Compaction commits were typed `pref-promote:` because that was the only type allowed to remove lines from `preferences.md`. Compaction is not promotion, and a log where both read `pref-promote` cannot tell a reader which happened.

**Discovery.** Nothing in the vendored `README.md` / `AGENTS.md` / `docs/` pointed at any of this. That pointer could only ever be written here.

## Gate calibration (#79)

From the null test in the ticket — two blind runs over the *same* rule set — which returned `pass` by luck:

| Problem | Evidence | Fix |
| --- | --- | --- |
| Unstable gated denominator | Both runs picked the same slot on all 17 cases but disagreed on whether a rule drove 4 of them, so n moved 3 → 5 under a non-change | `min_gated_cases` (default 8); below it the gate returns `insufficient-evidence`, accepted only with `replay_waiver_label` |
| Slot-ordering leak | `chosen_slot` was 1 in 14 of 17 records, so blind "always slot 1" scores 82%; both runs scored 94% | `cases` presents options in a per-record order derived from its ID; `score` derives the same order and maps back |

`fail` is never waivable — the waiver covers absent evidence, not a measured regression. The permutation mapping is never written into the cases file.

The remaining calibration is data, not code, and #79 leaves that box open deliberately.

## Incidental

`preferences_sha256` moved from `preferences_guard` to `replay`, so the dependency runs guard → replay one way only; the guard re-exports it.

Extraction's `--grep` pattern is written literally rather than through `re.escape`, which emits `pref\-extract:` — `\-` is undefined in the POSIX **basic** regex `git log --grep` uses.

## Verified

- `uv run pytest` — 115 passed.
- `prek run --all-files` — clean.
- Store self-test — **103 tests** (was 39), no skips when a corpus is present.
- **Against a real store's corpus** (17 records, scratch clone, discarded): guard clean, budget `~159 tokens (8.0% of 2000)`. All five watermark cases against real git history: bootstrap sees all 17; a record-adding branch fails without a pass; passes with one; fails when a record lands after the pass; the watermark then scopes the next batch to that one record.
- **Against two real draft batches**: 5 duplicates, 1 uncertain, 0 false positives.
- CLI exit codes: `gate` returns 0 / 1 / 3 for pass / fail / insufficient-evidence.

### Fixture store

The real-corpus checks used to skip here, so a template-side edit could ship a broken guard to every store that updates. `FixtureStoreTests` and `ExtractionWatermarkTests` build throwaway stores in a tempdir — including real git repos, so the watermark walk and the positional gate run against actual history rather than mocks.

### Not verified: `copier update` into a store repo

`copier update` replays the recorded `_commit` (`v3.1.0`) before diffing, so it cannot run against an unreleased branch. The render-manifest and file-ownership tests cover the same ground, and the repo's `conflict-detection` CI job runs the real update path. The real-store update happens after release, as a normal PR diff there.

## Constraint check

No store repo URL, name, owner, issue number or record slug appears in any file in this diff. The dry-run and draft-batch numbers quoted here and in `.github/store/README.md` are counts and rates only. Stdlib only throughout.


frankify-app:
Design change pushed in `2778bde`: **the extraction marker is gone.** The ticket body of #78 is updated to match.

The marker was a second watermarking scheme for a lifecycle that already had one. `record.py submit` computes `pref-confirm` counter bumps from the branch's records and commits them into the session PR — one of extraction's three outcomes, already PR-scoped and already marker-free. `git diff --diff-filter=A base...HEAD -- decisions/` answers "what is new" for the other two. The marker only added state to advance, keep in sync, and guard against going stale.

**Scope is not evidence**, which is what made per-session extraction look wrong. The pass ACTS on the records the PR adds and REASONS from the whole corpus, shipped in the batch as `history`. Cross-session repetition stays fully available — a pattern appearing once in this PR and twice in `history` is a three-record pattern. That was the property "batch, not per-session" was protecting, and it survives.

**The one thing the marker did buy is bought back explicitly.** A session PR merged without an extraction pass would otherwise leave no trace anywhere. So a PR adding records must carry an `<!-- extraction-pass -->` block naming exactly the records it adds, and the guard compares that list against the diff — a block copied from another branch, or covering half the records, fails rather than passing plausibly. A pass that found nothing declares zeroes; omitting the block is the only thing that fails, because silence is the failure mode a watermark existed to prevent.

`batch --all` scopes one pass to the whole corpus for records that predate extraction. A one-off, not a mode.

### What moved

| | before | after |
| --- | --- | --- |
| Batch boundary | `extraction-marker.json`, advanced in the PR's last commit | `git diff` against the PR base |
| Enforcement | guard checks the marker names a real record | guard checks the declared pass covers exactly the added records |
| Cross-session evidence | implicit in a multi-session batch | explicit `history` in every batch |
| Store-owned files | `preferences.md`, `store.config.json`, `extraction-marker.json` | `preferences.md`, `store.config.json` |

### Verified

`uv run pytest` 115 passed, `prek run --all-files` clean, store self-test 77 tests. End-to-end against a scratch clone of a real store with real git history: `status` reports 0-in-scope/17-in-history on an unchanged branch and 1/17 after a record lands; the guard rejects a record-adding PR with no pass block and accepts one whose block covers the added record.


---

frankify-app:
Second design change, pushed in `78b145e`: **the watermark is now derived from git history** — a `pref-extract:` commit — rather than from the PR diff. #78's body is updated; the PR description above is rewritten to match.

Third and last iteration on the same question: a marker file, then the PR diff, now git itself.

## Why the PR diff was not enough

It was correct but not *recoverable*. A session merged without a pass left no trace anywhere, so those records were gone — the CI gate mitigated it, but a gate bypassed once loses the records permanently. Walking back to the last pass commit means a missed session is simply picked up by the next pass. Recovery costs nothing and needs no state.

Two things fell out of deriving rather than storing:

- **Bootstrap stopped being a special case.** No `pref-extract:` commit anywhere means the watermark is the beginning of the corpus, so the first pass covers everything by construction. `--all` is deleted, not reimplemented.
- **The scope query and the guard check became the same concept**, so the `<!-- extraction-pass -->` block goes away entirely. The commit's *place* is its claim: nothing to enumerate, nothing to keep in sync with the diff, nothing copyable from another branch.

## The reliability question, answered

Detection only works because **an empty pass commits too**. A pass that finds nothing produces no proposal and no counter bump, so a watermark keyed on outcome commits would stall every time extraction legitimately had nothing to say. `pref-confirm:` is worse than useless for it — `record.py submit` emits those, so it would move the watermark for a pass that never ran, and a false watermark *skips* records rather than re-reading them. Hence a dedicated type nothing else produces, plus `--allow-empty`.

One implementation trap worth naming: the `--grep` pattern is written literally rather than through `re.escape`, which emits `pref\-extract:`. `\-` is undefined in the POSIX **basic** regex `git log --grep` uses, where backslash-punctuation is not portably a literal.

**Known constraint, now documented:** the repo must not squash-merge. Squashing rewrites subjects and the watermark is a subject in history. Merge commits and rebase both preserve it — this store merge-commits, so it holds today, but it is a repo setting that is invisible until it breaks.

## What moved

| | before | after |
| --- | --- | --- |
| Batch boundary | `git diff` against the PR base | last `pref-extract:` commit in history |
| Enforcement | `<!-- extraction-pass -->` block matching the diff | branch contains a pass commit, no record added after it |
| Missed session | lost, unless the gate caught it | picked up by the next pass |
| Bootstrap | `batch --all` | falls out — no commit means the whole corpus |

## Verified

`uv run pytest` 115 passed, `prek run --all-files` clean, store self-test 87 tests. End-to-end against a real store clone with real git history, all five cases:

1. no pass in history → bootstrap sees all 17 records
2. branch adds a record, no pass → guard **fails**
3. pass committed → guard **passes**, notes the watermark moving
4. record added *after* the pass → guard **fails**
5. watermark now scopes the next batch to that one late record

The store's bootstrap PR is updated to close with a real `pref-extract:` commit rather than a body block.


---

frankify-app:
Also closes #81 — the ingestion gate, pushed in `0c0b12b`. Folded in here rather than shipped separately because it is the same vendored directory and blocks the same draft batch.

`.github/store/similarity.py`, read-only, three checks in one run. They share a deadline: **drafts are mutable and records are not**, so anything not fixed at ingestion is frozen by append-only. That is what separates them from the analysis roadmap (#82), which reads immutable history and can wait for volume.

Classification is by **provenance**, which is what distinguishes one ruling recorded twice from the same question decided again later — same `preference_set.commit` + same `chosen` is a duplicate, distinct commits are a re-decision, an existing link outranks both. Chat drafts carry null provenance by design, so absence is never read as "distinct": it falls back to the session key, then the answer, and lands in `uncertain` rather than guessing.

Scoring uses question + option labels and deliberately **not** `chosen` — two records of one question with different answers are precisely the interesting case, and scoring the answer in would push them apart.

### Verified against the real 17-record corpus

```
ingestion gate: 17 candidate(s) — 0 duplicate, 0 re-decision, 0 uncertain, 1 linked

  [linked] 0.4455 +artifact  …rejection-reason-provenance  vs  …silent-pick-operative-reason

artifact_ref: 14 complete, 0 partial, 3 null
```

Zero false clusters, and the one genuinely related pair surfaced and correctly classified as **already linked** rather than duplicate — the case that would have been a false positive. It also found three null `artifact_ref`s, which is real actionable output. Synthetic duplicate / re-decision / linked / uncertain all classify correctly; false-cold flags a matching active rule and never fires on a preference-driven record.

Store self-test now 100 tests; template suite 115; `prek` clean.

### Scope note

#81 as filed bundles three tickets from the roadmap draft. Tickets 2 and 3 were originally proposed as "not needed yet", but they carry Ticket 1's exact urgency argument — the roadmap's own Eisenhower rule is "urgent = must act on mutable drafts (immutability makes delay permanent)". A false-cold stream or an unenriched ref is just as permanent as a duplicate record, and both are small additions to a report that was being generated anyway. The record-count-triggered passes stay parked in #82.


---

frankify-app:
Pushed `965d165` — the false-cold check was size-biased to the point of being structurally dead, found by running the gate over five real draft batches rather than fixtures.

Jaccard divides by the union. Rules run 7–8 tokens, records 20–42, so a 7-token rule quoted **verbatim** inside a long record maxed out at 0.167 against a 0.18 threshold. The check could never fire. Fixtures hid it because fixture records were short enough to stay under the size gap.

Now scores **rule coverage** — the share of the rule's terms present in the record. Size-invariant. Threshold 0.4, taken from the measured distribution on 77 real drafts + 17 records: one hit at 0.4286, a wide gap, then 0.25 and a floor at 0.14.

Rules are also stripped of their `[confirmed: N, last: DATE]` suffix before tokenizing — bookkeeping, identical across the set, and it diluted coverage by an amount that grew with how well-confirmed the rule was, so the best-established rules matched worst.

Surfaced one genuine candidate the old measure buried: `20260722T223525Z-adrs-coexist-with-decision-memory` at 0.4286 against the person-level/project-level rule.

Report key `overlap_score` → `rule_coverage`. Full validation numbers are on #81. 104 store tests, 115 template tests, all hooks green.

---

frankify-app:
Pushed `26626e6` — gate thresholds are now store-tunable and they expire. Full write-up on #84; the template-level distribution kata is #83.

Three parts, in the order they had to happen:

1. **The five gate thresholds move to `store.config.json`.** They were hardcoded in a vendored module, so a store that measured a better value had no legitimate way to apply it — editing vendored code is what the vendoring rule forbids, and changing the template constant pushes one store's calibration onto every consumer. Calibration is a property of a corpus.
2. **Calibration stamps + staleness.** Each constant carries `corpus_size`, `separation`, `measured`. Empty by default, so a store that has never measured says so rather than inheriting a stamp earned elsewhere. Reported by the gate as *never measured* or *outgrown* — a prompt, never a failing check, since failing CI over a stale threshold only teaches people to silence it.
3. **`/recalibrate-thresholds`** — measures each constant against the current corpus, finds the gap, proposes with the distribution attached, and never applies.

Plus three katas promoted from the real failures, and two of them failed on first run for good reasons: the "long" fixture was 4 tokens short of the real corpus maximum, and the override test raised one surfacing channel while leaving the other free to surface the pair.

Also adds `--out` to `similarity.py`. Two shipped skills already documented that flag and it did not exist, so the command in `/adjudicate-drafts` failed outright — now with a test so a documented flag cannot go missing again.

120 store tests, 115 template tests, all hooks green.

---

frankify-app:
**History consolidated** — force-pushed, 10 commits down to 6. The tree is byte-identical to the pre-rewrite state (verified by tree SHA), so nothing in the diff changed.

| | Commit | Folded in |
| --- | --- | --- |
| 1 | `feat: batch preference extraction from records` | — |
| 2 | `feat: close the store-layer seams and calibrate the gate` | — |
| 3 | `refactor: derive the extraction watermark from history` | the marker-file → PR-diff → git-history revisions, and the `--base` fixup |
| 4 | `feat: ingestion gate — dedup, re-decision links, false cold` | the real-data fixes (slug keying, answer comparison) |
| 5 | `feat: containment surfacing, adjudication, size-invariant false cold` | the false-cold rule-coverage repair |
| 6 | `feat: make gate thresholds store-tunable, and expire them` | — |

### One correction to this PR's description

The extraction section above says:

> Both earlier designs are gone, and the commits are in this branch's history rather than squashed away.

That is **no longer true** — the two superseded watermark designs are now folded into commit 3. Keeping commits whose design was then replaced is history-as-diary, which is what consolidation removes; the evolution and the reason each design was abandoned now live in that commit's message, which is the durable record. Flagging it rather than quietly leaving a false statement in the description.

120 store tests, 115 template tests, all hooks green after the rewrite.

---

decision-memory/.agents/skills/adjudicate-drafts/SKILL.md
```diff
@@ -0,0 +1,169 @@
+---
+name: adjudicate-drafts
+description: Work the ingestion gate's clusters into a decision the human can make in one pass — propose a resolution per cluster, lay out what each way costs, and name what becomes irreversible. Use after running the ingestion gate, when drafts are flagged duplicate/re-decision/uncertain, or when a drafts batch is about to be ingested.
```
frankify-app: Shorten to what this skill is for, not how it works. Skill token budget in descriptions is very scarce

decision-memory/.agents/skills/adjudicate-drafts/SKILL.md
```diff
@@ -0,0 +1,169 @@
...
+  repo and path beats a fabricated commit.
+
+## Afterwards
+
+Ingest through the recorder as normal. The records are immutable from
+that moment, which is the whole reason this pass exists.
+
+If adjudicating produced a ruling worth remembering — a policy on when
+two extractions count as one decision, say — that is a decision like
+any other and belongs in a grilling session, not in this pass.
```
frankify-app: Condense the writing of the whole skill. You don't need to convince the reading agent thy the skill is correct, things like "Why the deadline is real." do not add meaningful information. Use caveman mode and reduce it to its substance

frankify-app: It is fine to give a hint _why_ the instruction is there when needed or to give an example, but no flowery prose to pitch the idea or pitch the pitch

frankify-app: Apply this also for other skills or md files here. If this is not already in AGENTS or other files we should fix that in the template repo -> ticket if needed

decision-memory/.agents/skills/adjudicate-drafts/SKILL.md
frankify-app: Same for other skills

decision-memory/.github/store/README.md
```diff
@@ -43,20 +48,64 @@ these knobs:
...
+| `replay_waiver_label` | `preferences-replay-waiver` | label accepting an `insufficient-evidence` gate |
 | `replay_window` | 20 | how many recent decisions the replay scores |
+| `min_gated_cases` | 8 | below this many preference-driven cases the gate reports `insufficient-evidence` |
+| `similarity_threshold` | 0.35 | below this a pair is not worth a human's attention |
+| `containment_threshold` | 0.5 | the split channel: one draft re-extracted as two |
+| `artifact_boost` | 0.15 | how much repo+path agreement lifts a borderline pair |
+| `answer_agreement` | 0.5 | how close two rewordings must be to count as one answer |
+| `false_cold_threshold` | 0.4 | rule coverage that flags a suspect `cold` record |
+| `calibration_growth_factor` | 2.0 | corpus growth past a stamp before it reads as stale |
+| `calibration` | `{}` | evidence behind each calibrated constant |
```
frankify-app: If we name the concrete numbers here this is prone to drift if numbers change, rather refer to the config and remove that column

decision-memory/docs/conventions.md
```diff
@@ -183,9 +183,11 @@ everywhere — new optional fields need no migration.
 
 ## Active preference set (`preferences.md`)
 
-- Hard token budget: ~2k tokens, CI-enforced. Promoting a rule at
-  budget means merging or demoting another ("promote requires
-  demote").
+- Hard token budget, CI-enforced. The number is `budget_tokens` in the
+  store-owned `store.config.json` (default ~2k); the vendored guard
```
frankify-app: Don't need to name the default number explicitly here (drift)

decision-memory/docs/conventions.md
```diff
@@ -211,15 +287,42 @@ commit:
 - `pref-proposal: <rule>`
 - `pref-promote: <rule>` (human only)
```
frankify-app, line 288: I guess the promotion suggestions will be done by the agent, but gated by the human, right? Because otherwise it will be too much work to extract promotion candidates by hand

decision-memory/AGENTS.md
```diff
@@ -41,17 +46,78 @@ authoritative contract.
...
   [docs/conventions.md](docs/conventions.md) (`decision(...)`,
-  `pref-proposal`, `pref-promote`, `pref-confirm`, `chore`).
+  `pref-proposal`, `pref-promote`, `pref-confirm`, `pref-compact`,
+  `pref-drift`, `pref-extract`, `chore`).
 - In managed environments (e.g. Claude Code on the Web), ALWAYS use
   the tooling the environment itself declares for forge operations
   (PR creation etc.) — `gh`/`curl` are typically sabotaged there. The
   recorder's `submit` hands the PR off to you in that case.
 
+## Before ingesting drafts
```
frankify-app, line 56: Do we say that drafts need to be split into individual record files and individual commits for each file?

decision-memory/docs/conventions.md
frankify-app: Please also check for other numbers that are adjustable and might drift in the future and rather refer to where they are configured

frankify-app:
Review addressed in `9553076`. Two of the ten comments were questions — answered below rather than changed, except where the docs said it in the wrong place.

## Skill prose and descriptions

Descriptions now say what the skill is for and when to trigger it, nothing about mechanism:

| Skill | Description |
| --- | --- |
| `adjudicate-drafts` | 338 → 183 chars |
| `compact-preferences` | 383 → 231 |
| `extract-preferences` | 444 → 214 |
| `recalibrate-thresholds` | 353 → 213 |

Bodies: ~110 lines out across the four, no procedural content dropped. Gone are the headings that argued the skill was correct — "Why the deadline is real", "Why it is not cosmetic" — and the restatements around them.

I kept reasons in the few places where the reason changes what the agent does. The clearest example is in `adjudicate-drafts`: a duplicate is not just a wasted file, it manufactures the cross-record repetition that extraction reads as evidence a pattern is a principle. An agent that does not know that will rank a duplicate as cosmetic. Three sentences, and they change the decision — that is the line I drew against "no pitch".

## Configurable values no longer repeated in docs

- **Store README config table:** the `default` column is gone entirely. `config.DEFAULTS` is the one authority; the table now describes meaning only.
- Swept the rest as asked, and found four more beyond the one flagged: the inline `~2k` budget in `README.md`, `conventions.md` and the seeded `preferences.md`, the `~20` replay window in `compact-preferences`, and two literal label names (`preferences-carve-out`, `preferences-replay-waiver`) in that same skill. All now refer to their config key.
- Verified nothing is left: every remaining mention is a key name, never a value.

## The two questions

**Promotion candidates — agent proposes, human gates?** Yes, exactly that, and it is already the documented contract. The agent's extraction pass writes candidates to `proposals/` (one rule per file) and flags drift as a proposal file; only a human `pref-promote` commit moves text into `preferences.md`. `conventions.md` states "merging a proposal file is NOT promotion", `AGENTS.md` states "promotion is human-only, always", and the commit-type list marks `pref-promote` human-only. The agent never has promotion power — extracting candidates by hand was never the intent.

**Do we say drafts split into individual files and commits?** It was said, but in the wrong place — `AGENTS.md` § Git, `conventions.md` § PR flow, and the README write-flow bullet, none of which is where someone ingesting a batch is reading. Added it to `AGENTS.md` where drafts are actually discussed, including *why* it matters: one file plus one commit per draft is what makes partial acceptance possible, since a reviewer drops or reverts individual commits before merge.

## Ticket filed

**#85** — a skill-authoring convention for `AGENTS.md`. There is currently no such guidance anywhere in the repo, so nothing would have caught this; the four skills here conform now, but the next one written would drift the same way. Filed at template level rather than in the store subtemplate because `.agents/skills/` is template-wide.

120 store tests, 115 template tests, all hooks green.

---

decision-memory/docs/conventions.md
frankify-app: Yes — agent proposes, human gates, and that is already the contract rather than something this PR changes.

The extraction pass writes candidates to `proposals/` (one rule per file) and flags drift as a proposal file too. Only a human `pref-promote` commit moves text into `preferences.md`. The agent's only write to the active set is a `pref-confirm` counter bump, whose arithmetic CI validates.

Stated in three places: `conventions.md` — "merging a proposal file is NOT promotion"; `AGENTS.md` — "promotion is human-only, always"; and the commit-type list marks `pref-promote` human-only.

So extracting promotion candidates by hand was never the intent — the agent does the reading and drafting across the whole corpus, you decide what gets adopted.

No change made for this comment.

---

decision-memory/AGENTS.md
frankify-app: We did say it — but in three places nobody ingesting a batch is reading: `AGENTS.md` § Git ("one PR per session; one commit per record"), `conventions.md` § PR flow ("ONE commit per record — atomic and dissectable"), and a README write-flow bullet ("one immutable JSON file per decision").

None of those is here, where drafts are actually being handled, so I added it to this section in `9553076` — including why it matters:

> Then ingest. Each surviving draft becomes **one record file and one commit** — `record.py record` does the split, so a batch is never one bulk commit. That is what makes partial acceptance possible: a reviewer drops or reverts individual commits before merge.

The recorder already enforces it, so this is documentation catching up with behavior rather than a new rule.

---

Issue #85, opened 2026-07-27
Write a skill-authoring convention into AGENTS.md (description budget, no pitch prose)


frankify-app:
From review feedback on #80. The fixes landed for the four decision-memory skills, but there is **no convention anywhere that would have prevented it**, so the next skill written in this repo — or in any stamped repo — will drift the same way.

`AGENTS.md` § Skills currently covers loading, syncing and the trigger table. It says nothing about how to *write* one.

## What the convention should say

**Descriptions are the scarcest budget.** A skill's `description` loads whether or not the skill ever runs, so it pays rent on every session. It should say what the skill is FOR and when to trigger it — not how it works. The bodies loaded in #80 ran 338–444 characters describing mechanism; the rewritten ones run 183–231 saying purpose and trigger.

**Bodies instruct, they do not persuade.** The reviewer's phrasing is the rule worth adopting verbatim:

> It is fine to give a hint *why* the instruction is there when needed or to give an example, but no flowery prose to pitch the idea or pitch the pitch.

The failure mode has a recognisable shape: section headings like "Why the deadline is real" and "Why it is not cosmetic" that argue the skill is correct to a reader who has already loaded it and is trying to follow it. A one-line reason earns its place when it changes what the agent does; a paragraph defending the design does not.

## Why it belongs in the template rather than in the store subtemplate

Skills are a template-wide artifact — `.agents/skills/` exists in every stamped repo, and the trigger tables in `AGENTS.md` are template-level. A convention written only into the decision-memory subtemplate would not reach any other skill.

## Acceptance

- `AGENTS.md` § Skills gains a short authoring subsection: description = purpose + trigger, bodies = instructions with reasons only where they change behavior, no argumentative headings.
- Stated tightly enough to obey its own rule — a long section about writing short sections would be self-refuting.
- Optionally a mechanical check (description length ceiling) if it can be done without a dependency; worth a look, not worth blocking on.

## Related

The four decision-memory skills already conform, in `9553076` on #80 — usable as the reference for what "conforming" looks like.

frankify-app:
Heads-up before anyone implements this as written: frankify-app/skills#29 derives a `writing-skills` skill that owns skill-authoring end to end, and it was grilled to **single home** explicitly so the rules live in one place. Executing this ticket's acceptance criteria unchanged would create a second home and put us on the wrong side of § Avoid Duplication.

[...]

PR #88, opened 2026-07-27
docs(decision-memory): add the pre-merge link pass to adjudicate-drafts

### LINKED ISSUE #87: Ingestion gate: identical slugs never surface, and it cannot see records in open PRs
> **Note:** this body is the authoritative version. Two earlier comments on this issue contain store-private detail and are superseded by the text below.

The gate scores `question` + option labels and **never looks at the slug**, so two records on one subject with an identical slug can score below both channels and never surface. `identity()` reads the slug for *reporting*; nothing reads it for *scoring*.

## The case, in the abstract

Two records about the same subject, one on `main` and one on an open PR. Same slug — slugs follow the topic, so two rulings on one topic collide naturally. Same final answer, because the answer held between the two rulings.

They are **not duplicates**. Reading both in full:

| | A | B |
| --- | --- | --- |
| context | the decider arrived with the answer already chosen | a full exploration with a shortlist |
| options | 1 | 4 |
| chosen_slot | 1 | 4 — free text, beyond the listed options |
| rejections | none | 5, one operative |
| outcome | hit | refined |

Two rulings on one subject, very different deliberation. Both belong in the corpus.

Their scores:

```
jaccard      0.08   (similarity_threshold 0.35)
containment  0.40   (containment_threshold 0.50)
```

Below both channels — **invisible**.

## Why that matters, precisely

Not because a duplicate would have been ingested. Because these two want a **link** and have none. A re-decision that lands unlinked is a permanently disconnected node, which is one of the three things this gate exists to prevent.

B already carries `related` edges to two sibling records. It has none to A, because A was invisible: A lives in an open PR.

## Wrong fix, and why

An earlier version of this ticket proposed classifying an exact slug match as `duplicate`. That is **wrong** — this pair is a live counter-example. Auto-classifying on slug would have told a human to discard a record carrying five rejections and a `refined` outcome.

An identical slug should surface as **`uncertain`**, never as `duplicate`.

## The same record produced four false positives

A short record (5 content tokens) paired against four unrelated store records at containment 0.6–0.8, purely because most of a 5-token set is contained in anything sharing two of its words.

So for one record: **four false positives and one false negative**, and the false negative is the one that mattered. Containment is size-invariant by construction, which is what makes it good at splits — but below some size the ratio stops meaning anything.

## Cross-PR references are impossible, so the fix is discovery

`decision_validator.validate_corpus` requires every link target to exist:

```python
if ref not in records:
    errors.append(f"{record_id}: {field} points to {ref!r}, "
                  "which does not exist in decisions/")
```

Two PRs open at once each lack the other's records, so any edge between them is a dangling ref and fails CI. That is correct — a record must not point at something that may never merge.

What follows:

- The edge can only be written by the PR that merges **second**, pointing back at the first.
- Once the first merges, its records are immutable, so it can never gain the reverse edge.
- Cross-PR relationships are therefore permanently **one-directional**.

None of that should change. What is broken is *when* the need is discovered: the gate cannot see the pair while both PRs are open, and the moment it becomes writable — the first PR merging — nobody is looking. **The knowledge expires exactly when the ability to act appears.**

## The fix

A **pre-merge link pass**: immediately before merging, after rebasing onto current `main`, re-run the gate over the branch's records against the now-complete corpus and write any links it surfaces.

This is cheap and already sanctioned:

- `check_append_only` uses a three-dot diff (`base...HEAD`), so a record added and later amended in the same PR reads as `A`. An ordinary follow-up commit works; no history rewriting.
- `docs/conventions.md`, PR flow: *"Partial accept = hand-edit the branch, drop or revert individual commits before merge."* Records are not immutable until merge — merging is the acceptance.

**And it makes the open-PR problem disappear.** For any two records that want an edge, the PR merging second sees the first, because by then it is merged. If every PR runs this pass right before merging, the later one always catches it — no cross-PR machinery needed, which is strictly simpler.

## Acceptance

- Two records with the same slug surface as `uncertain` — a prompt to consider a link, never an instruction to discard.
- Containment does not fire on records too small for the ratio to mean anything.
- A pre-merge link pass is documented in the skill, with the narrower remedy set it has (link, or drop the commit — there is no drafts file to discard into).
- Regression test built from this pair's *shape*: same slug, same answer, different decisions — must surface, must not read `duplicate`.

## Gaps this exposes

- **No commit type fits** a commit that adds a link to an unmerged record. `decision(...)`, the `pref-*` family and `chore:` are the options; `chore:` works but says nothing.
- **Read-time asymmetry** (noted, not proposed): because edges only point backwards, reading the earlier record alone never reveals that a later ruling refined it. Solvable by scanning for inbound edges at query time, so it does not belong here.

## Related

#84 — the containment half is a calibration question. #86 — same family: a signal present in the data but unused, or used without validation.

frankify-app:
Closes the discovery half of #87.

## The two gaps

The gate's remedies were all drafts-side, which left records unreachable in two situations:

1. **A session that records natively** through the recorder never produces drafts, so it never ran the gate at all.
2. **Even a batch that did run it** was gated against the corpus as it stood at record time. Other PRs land in between, so the check was already stale by the time the PR merged.

Both were found on real data: two records on one subject, one merged and one in an open PR, with no link between them and no moment at which anything would have proposed one.

## The pass

Runs on **every** record-adding PR, immediately before merge and after rebasing — the last moment the corpus is complete and the branch is still mutable.

That window exists and is already sanctioned:

- `check_append_only` diffs `base...HEAD`, so a record added and later amended within one PR still reads as an addition. An ordinary follow-up commit works; no history rewriting.
- `docs/conventions.md` § PR flow: *"Partial accept = hand-edit the branch, drop or revert individual commits before merge."* Merging **is** the acceptance.

## The ordering constraint is documented, not worked around

A link target must already exist in `decisions/` — the validator rejects a dangling ref, correctly, since a record must not point at something that may never merge.

So two open PRs cannot reference each other. The edge is written by **whichever merges second**, pointing back at the first, and the first can never gain the reverse edge.

**Applied uniformly that is sufficient.** The later PR always sees the earlier one, so nothing needs to read across open PRs — which is why this PR closes #87's discovery problem without building the cross-PR machinery that ticket originally proposed.

## Narrower remedy set, stated plainly

| Remedy | Available |
| --- | --- |
| Add a link | yes — the point of the pass |
| Drop the record | only by dropping its commit |
| Discard to `discarded-drafts.json` | no — drafts-only |
| Edit the other side | never — merged and immutable |

## A shared slug is not evidence of a duplicate

Documented explicitly, because the real pair that motivated this looks exactly like a duplicate and is not: same slug, same final answer, but one record has a single option and no rejections while the other has four options, five rejections and a `refined` outcome.

Slugs follow the topic, so two rulings on one topic collide naturally. That pair scored **0.08 jaccard / 0.40 containment** — below both channels. The scoring fix stays in #87; this documents the judgement so nobody discards a record on a slug match.

## Description broadened

A natively-recorded session would not otherwise load this skill — the old description only mentioned drafts.

## Verified

115 template tests, 120 store tests, all hooks green.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MJ7c5rzeXMCMvZTofawmwn)_

### DIFF
--- a/decision-memory/.agents/skills/adjudicate-drafts/SKILL.md
+++ b/decision-memory/.agents/skills/adjudicate-drafts/SKILL.md
@@ -1,9 +1,9 @@
 ---
 name: adjudicate-drafts
-description: Resolve the ingestion gate's flagged draft clusters before ingestion. Use when drafts are flagged duplicate, re-decision or uncertain, or before ingesting a drafts batch.
+description: Resolve the ingestion gate's flagged clusters — in drafts before ingestion, in a branch's records before merge. Use when drafts are flagged duplicate/re-decision/uncertain, or before merging any PR that adds decision records.
 ---
 
-# Adjudicating flagged drafts before ingestion
+# Adjudicating flagged records
 
 > Copier-vendored from the agentic-engineering-template decision-memory
 > subtemplate — do NOT edit it in the store repo;
@@ -11,6 +11,14 @@ description: Resolve the ingestion gate's flagged draft clusters before ingestio
 
 The gate finds clusters and does not resolve them. This skill turns
 each cluster into a decision the human answers in one pass.
+
+Two moments, and every record-adding PR hits at least the second:
+
+1. **Drafts, before ingestion** — sections 1-5. The full remedy set is
+   available because nothing is written yet.
+2. **A branch's records, before merge** — the pre-merge link pass at
+   the end. A session that recorded natively through the recorder
+   never had drafts and starts there.
 
 `decisions/` is append-only with no carve-out: after ingestion a
 duplicate cannot be withdrawn and a missing link cannot be added.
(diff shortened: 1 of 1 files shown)


decision-memory/.agents/skills/adjudicate-drafts/SKILL.md
```diff
@@ -124,9 +132,88 @@ Then, while the drafts are still mutable:
...
+  outcome; record that you looked.
+
+Two records sharing a slug are worth reading closely and are **not**
+thereby duplicates. Slugs follow the topic, so two rulings on one topic
+collide naturally — the deliberation tells you which it is: option
+count, rejection depth, and whether the outcome was a plain hit or a
+refinement.
+
+Commit the links as `chore:` — they are not new decisions, and no other
+commit type fits.
```
frankify-app: No pitching needed, reduce to necessary instructions only, no flowery prose. use caveman mode

frankify-app (commented):
Don't forget to reapply template to self at the end

decision-memory/.agents/skills/adjudicate-drafts/SKILL.md
frankify-app: Cut in `364981e` — **88 lines to 51**.

Gone: the paragraph arguing why the timing is right, the one explaining why the pass runs on every PR, and the framing about how this judgement differs from the drafts case. All of it defended the design to a reader who has already loaded the skill and is trying to follow it.

Kept two reasons, because each changes what the agent does rather than justifying the design:

- *"Rebase first, or the gate checks against a corpus missing everything merged since the branch was cut"* — without it, the pass runs and silently checks nothing useful.
- *"A shared slug is not a duplicate"* — without it, the agent discards a record on a slug match, which is the exact mistake I made on the real pair.

Section reads as command, table, constraint, judgement list now.

Worth noting this is the second time on the same lesson — I condensed all four skills for this in `9553076` and then wrote 88 lines of new prose two hours later. #85 (the skill-authoring convention for `AGENTS.md`) is the thing that would actually stop it recurring, since nothing in the repo currently states the rule.

---

PR #94, opened 2026-07-28
refactor: extract the shared record cores so a second store can compose on them

hvoecking:


pando-genet:
Closed: opened from the wrong account. Recreated verbatim as #97 from the same branch.

---

Issue #115, opened 2026-07-29
record.py submit bumps preference counters without the rule_driven_acceptances exclusion


pando-genet:
## The defect

Two paths write preference counters. Only one applies the exclusion that makes
them mean anything.

| Path | Bumps counters | Honours `rule_driven_acceptances` |
| --- | --- | --- |
| `extract-preferences` skill | yes | **yes** |
| `tools/record.py submit` | yes | **no** |

The rule `submit` is missing, from the extraction skill:

> Do **not** bump on a record in `rule_driven_acceptances` — there the rule cited
> itself into the prediction slot and that slot was chosen, so the recommendation
> caused the choice it would be credited with predicting. Zero independent
> evidence.

## Observed, not inferred

In `pandoscope/decision-memory`, the active rule

> Scopes work to the demonstrated problem … `[confirmed: 4, last: 2026-07-28]`

reached 4 via four `pref-confirm:` commits written by `submit` in a single
session on 2026-07-28. The extraction pass run today
(`pref-extract: 19 records`) classifies **every one of those records** as a
rule-driven acceptance:

```
20260728T214205Z-evidence-store-bare-minimum-first
20260728T214205Z-shared-cores-copied-across-subtemplates
20260728T214205Z-sibling-writer-on-shared-core
20260728T214205Z-tier-boundary-is-runnability
20260728T214205Z-tier2-gate-mechanical-two-lane
```

Under the extraction rules that counter is **0**, not 4.

All five prior `pref-extract:` passes in that store report `0 confirmed`, so the
exclusion has never been applied to a single bump in the store's history. Every
counter in `preferences.md` came from `submit`.

## Why it is not cosmetic

The counter is the visible evidence a rule earns its context budget, and grilling
cites rules by strength. A high counter gets the rule cited into slot 1; the slot
is chosen; `submit` counts it. **A closed loop with no external input.**

It has already produced a wrong prediction: `20260729T073614Z-stores-split-by-layer-not-forked`
is a preference-driven miss where that rule lost to a structural-split ruling —
the first time it met a case its own recommendation had not framed.

The same loop explains the 5/5 preference-driven hit rate in decision-memory PR
#1, which the skill itself names as a smell rather than a result.

## Fix

`submit` applies the same exclusion, computed the same way, or stops bumping
counters entirely and leaves that to extraction.

The second is worth considering seriously and may be the better answer: the
extraction pass is where the corpus is read as a whole, where `history` is
[...]

PR #117, opened 2026-07-29
feat(decision-memory): count self-credited and independent confirmations apart

pando-genet:
Closes the writer half of #115.

## The defect

`confirmed` could only ever count a rule agreeing with itself. The
writer bumps on a preference-driven `hit`, and a `hit` means the slot
the rule was cited into is the slot the decider chose. So the counter
credited a rule for predicting its own recommendation.

Every bump in the store's history took that path. That is why five
extraction passes in a row reported `0 confirmed` against rules the
active set displayed at `n=4` — the pass excludes rule-driven
acceptances, and the writer never has. The two numbers were measuring
different things and neither said so.

## The fix

`submit` now records both kinds and keeps them apart:

- a `hit` raises `confirmed` alone — the rule agreeing with the option
  it authored
- a rule cited on an option the rule did **not** propose, which the
  decider chose anyway, raises `independent` as well

Only the second is evidence a rule tracks the decider rather than
steering them.

## Why the grammar had to change first

`conventions.md` declared the suffix as flat, order-insensitive and
tolerant of unknown keys. The code was `\[confirmed: (\d+), last:
(\d{4}-\d{2}-\d{2})\]` — one regex pinning both the key set and their
order. The documented format was not implemented, and adding a key is
what surfaced it. Replaced by a `parse_metadata`/`format_metadata` pair
in the validator, which is the single source the guard and the writer
both consume.

## `independent` is required

The first draft made it optional, omitted meaning "unmeasured, which is
not the same as zero." A real distinction — with no instance behind it.
The only bump path could not produce anything but zero, so there was
never an unmeasured rule for the optional form to describe. It cost a
branch in every reader to represent a state that cannot arise. The key
is required; rules predating it are backfilled with the zero they had
measured all along.

## Guard changes

- counter math goes through the parser, not a regex
- a bump may move `confirmed`, `last` and `independent` only;
  `independent` rises by at most 1 and never falls
- every other key is frozen, so a rewrite hiding behind a bump's commit
  subject fails instead of merging as routine
- the corpus check now verifies every rule bullet carries its counters,
  so a rule that loses them fails rather than reading as one nobody has
  ever confirmed

## Tests

201 passed, 4 skipped. The grammar tests go red against the old regex.

One test was written and then removed: asserting the template's own
`preferences.md` has valid suffixes passes vacuously, because the
template ships the seed and the seed has no rules. It would stay green
for as long as the file stayed empty. Replaced with a fixture that
actually contains a broken rule.

---
_Generated by [Claude Code](https://claude.ai/code/session_015EAmu4EbQ2s3nw4Q1aNVWz)_

### DIFF
--- a/decision-memory/.github/guards/decision_validator.py
+++ b/decision-memory/.github/guards/decision_validator.py
@@ -76,10 +76,100 @@ MAX_SLUG_LENGTH = validator_core.MAX_SLUG_LENGTH
 ID_RE = validator_core.ID_RE
 DATE_RE = validator_core.DATE_RE
 
-# Single source for the preferences counter-line grammar
-# ("[confirmed: N, last: YYYY-MM-DD]"): the guard's counter-math check
-# and the writer's pref-confirm bumps both consume this regex.
-COUNTER_RE = re.compile(r"\[confirmed: (\d+), last: (\d{4}-\d{2}-\d{2})\]")
+# Single source for the preferences metadata-suffix grammar (see
+# decision-memory/docs/conventions.md): one trailing bracket holding
+# comma-separated `key: value` pairs. The guard's counter-math check
+# and the writer's pref-confirm bumps both consume these.
+#
+# Parsed rather than pattern-matched so the key set is one tuple a
+# reader can check against the doc, not a shape encoded in a regex.
+METADATA_RE = re.compile(r"\[([^\]\[]*)\]\s*$")
+
+COUNTER_KEY = "confirmed"
+INDEPENDENT_KEY = "independent"
+DATE_KEY = "last"
+
+# Exactly these, on every rule, in this order. A closed set is what
+# keeps the suffix from gaining keys nothing reads: adding one is a
+# deliberate change here, with its consumer, rather than something a
+# writer can introduce in passing.
+SUFFIX_KEYS = (COUNTER_KEY, INDEPENDENT_KEY, DATE_KEY)
+
+
+def parse_metadata(line: str) -> dict[str, str] | None:
+    """Parse a rule line's trailing metadata bracket.
+
+    Returns the key/value pairs, or None when the line carries no
+    bracket or the bracket is not this grammar (a bare `[note]`, a
+    markdown link). Values are returned as written; callers coerce.
+    """
+    match = METADATA_RE.search(line)
+    if not match:
+        return None
+    pairs: dict[str, str] = {}
+    for chunk in match.group(1).split(","):
+        key, sep, value = chunk.partition(":")
+        if not sep:
+            return None
+        pairs[key.strip()] = value.strip()
+    return pairs or None
+
+
+def format_metadata(pairs: dict[str, str]) -> str:
+    """Render pairs back into a suffix, in SUFFIX_KEYS order."""
+    body = ", ".join(f"{k}: {pairs[k]}" for k in SUFFIX_KEYS if k in pairs)
+    return f"[{body}]"
+
+
+def strip_metadata(line: str) -> str:
+    """The rule text alone — what two rules are compared on."""
+    return METADATA_RE.sub("", line).rstrip()
+
+
+def counter_of(line: str) -> int | None:
+    """The `confirmed` value of a rule line, if it has a valid one."""
+    pairs = parse_metadata(line)
+    if not pairs or COUNTER_KEY not in pairs:
+        return None
+    try:
+        return int(pairs[COUNTER_KEY])
+    except ValueError:
+        return None
+
+
+def check_metadata_suffix(line: str) -> str | None:
+    """Return an error for a rule bullet whose suffix is malformed.
+
+    Applied to every `- ` bullet in preferences.md, so a rule that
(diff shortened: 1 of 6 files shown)


decision-memory/.github/guards/decision_validator.py
```diff
@@ -76,10 +76,113 @@
...
+# Single source for the preferences metadata-suffix grammar (see
+# decision-memory/docs/conventions.md): one trailing bracket holding
+# comma-separated `key: value` pairs. The guard's counter-math check
+# and the writer's pref-confirm bumps both consume these.
+#
+# Deliberately parsed rather than pattern-matched. The old regex pinned
+# both the key set and their order, so `conventions.md` could declare
+# "unknown keys are tolerated, order is not significant" while the code
+# rejected any such line — a doc that describes a format nothing
+# implements. Adding one key is what surfaced it.
```
pando-genet: No need to mention what old code did, that would only go into decision memory if at all

decision-memory/.github/guards/decision_validator.py
```diff
@@ -76,10 +76,113 @@
...
+
+# Kept for the counter-math check, which still needs to locate the
+# `confirmed` value specifically. Everything structural goes through
+# parse_metadata.
+COUNTER_KEY = "confirmed"
+INDEPENDENT_KEY = "independent"
+DATE_KEY = "last"
+
+# Every rule carries all three. An earlier draft let `independent` be
+# omitted to mean "unmeasured, which is not the same as zero" — a
```
pando-genet: Same here, no one will care what was done in a draft. We should make it a rule that comments should only state the current situation, and only if required the future, but never the past (that can also confuse agents)

pando-genet: What would be ok is to name the reason why the choice was made, but don't explain how it was before I would say

decision-memory/docs/conventions.md
```diff
@@ -197,17 +197,19 @@ One bracket at the end of a rule bullet, holding comma-separated
...
 
 Grammar, and it is deliberately this small:
 
 - **Flat only.** No nesting, no lists, no objects. A value is a bare
   scalar, so the whole suffix parses by splitting on `,` then `:`.
-- **Order is not significant**, but `last` reads better trailing.
-- **Unknown keys are tolerated**, so a key can be added without a
-  coordinated change to every reader.
+- **Order is not significant**, but the canonical write order is
+  `confirmed, independent, last`.
```
pando-genet: If it is not significant we don't need to explicitly say that. Since we have examples and precedence in other files agents will likely stick to that order without this instruction

pando-genet: Learning: don't say what is not required, unless it makes a solution simpler (which is not the case here I'd say)

decision-memory/docs/conventions.md
```diff
@@ -197,17 +197,19 @@ One bracket at the end of a rule bullet, holding comma-separated
...
+- **All three keys are required.** A rule without them fails the guard.
+- **Unknown keys are tolerated** and survive a counter bump untouched,
+  so a key can be added without a coordinated change to every reader.
+  A bump that *changes* one is not a bump — it needs the carve-out.
```
pando-genet: Let's not allow unknown keys, that just encourages drift, maybe let's rather enforce that the exact set of keys is present

decision-memory/docs/conventions.md
```diff
@@ -216,9 +218,19 @@ Keys currently defined:
...
 
+`independent` is required rather than optional, and that is a
+correction. It was first specified as omittable, meaning "unmeasured,
+which is not the same as zero" — a real distinction with no instance
+behind it. The writer's only bump path credits a rule for predicting
+the slot it was cited into, so it could not produce anything but zero;
+there was never an unmeasured rule for the optional form to describe.
+Representing a state that cannot arise cost a branch in every reader,
+so the key is required and the rules that predated it were backfilled
+with the zero they had measured all along.
```
pando-genet: Remove this, not relevant, all keys are required anyways

decision-memory/docs/conventions.md
pando-genet: Done — the set is closed. `SUFFIX_KEYS` is exactly `confirmed,
independent, last`, and `check_metadata_suffix` fails on anything
missing or extra.

This turned out to simplify rather than cost. Tolerating unknown keys
required the guard to freeze them across a bump, so a rewrite could not
hide behind a mechanical commit subject — about 25 lines that exist
only to police keys nobody reads. Closing the set deleted all of it;
the remaining check is just the `independent` bounds.

Two tests went with it (`unknown keys survive a bump`, `a bump may not
rewrite an unknown key`), replaced by one asserting an unknown key is
rejected.

The comment-hygiene point I applied across the whole diff, not only the
flagged lines — including one in `record.py` that explained what the
writer "used to" record. And I dropped the order-is-not-significant
line: the examples carry the order, and stating a non-requirement adds
a rule without simplifying anything.

Pushed as d0362f8.

---

PR #131, opened 2026-08-04
Every repo reports its ticket closes to the session-memory store

pando-genet:
Refs pandoscope/session-memory#2, pandoscope/session-memory#3.

The store closes a thread when the thread's own ticket closes. Its first design polled every live ticket on a six-hour clock; the principal ruled that unacceptable (decision-memory: `close-loop-trigger-poll-overruled`), and the correction is this: the template ships an `issues: closed` workflow to every stamped repo, which fires `repository_dispatch` at the store the moment a ticket closes. Event-driven end to end, on the same channel that already ships the labels and board workflows.

## Shape

- **`ticket-closed.yml`**, GitHub-forge only, beside `labels.yml` and `add-to-project.yml`. Adopted at root too — this repo mints tickets that ledger threads reference, and the self-application gate requires it anyway.
- **The store's location is `vars.SESSION_MEMORY_REPO`**, an org-level variable, deliberately not baked in at stamping time: the rendered file names no org's store, and a store move is one variable edit rather than a re-stamp of every repo. Unset → the job shows SKIPPED, never a silent green (same rule as the labels guard).
- **The sender signs as the release-bot app** (`RELEASE_BOT_CLIENT_ID` + `RELEASE_BOT_PRIVATE_KEY`, the exact pattern of `template-update.yml`), at the principal's direction: a fine-grained PAT's mandatory one-year expiry means a yearly morning where every sender in the org goes red at once, and the app's key never expires. Reuse over a second app because the key already circulates to every stamped repo — this PR introduces **no new secret**. The minted installation token is scoped to the store repository alone, so the sender never runs with more reach than the one call it makes.
- **The payload carries only the ticket's name, and the receiver ignores even that.** The store re-reads ticket state itself, so a stale or replayed dispatch is a wasted lookup, never a wrong event — there is no parser at the trust boundary at all (the concern AET#111 documents).

## Setup this needs, once

1. `SESSION_MEMORY_REPO` org variable = the store's `owner/repo`.
2. Release-bot app: no permission change for the sender (Contents write suffices for `repository_dispatch`). The **receiver** side wants Issues: Read-only added — detailed in pandoscope/session-memory#3.

## Testing

`test_github_forge_ships_ticket_close_dispatch` pins the dispatch call, the app-token mint, the store-only token scope, the SKIPPED guard, and — separately — that the rendered file names no org's store. The e2e expected-tree manifests gained the file. Full suite: 206 passed, 4 skipped.

## Rollout

Stamped repos pick the sender up on their next template update. Until a repo carries it, its closes reach the ledger via the store's manual sweep (`workflow_dispatch` on close-loop), which reconciles every live ticket — the receiver side is pandoscope/session-memory#3.

### DIFF
--- a/.github/workflows/ticket-closed.yml
+++ b/.github/workflows/ticket-closed.yml
@@ -1,0 +1,62 @@
+name: Report ticket close to the ledger
+
+# The session-memory store closes a thread when the thread's own ticket
+# closes. The store cannot hear this repository's events — a workflow
+# only receives events from the repository it sits in — so every repo
+# that mints tickets reports its own closes.
+
+on:
+  issues:
+    types: [closed]
+
+permissions: {}
+
+concurrency:
+  group: ticket-closed-${{ github.event.issue.number }}
+
+jobs:
+  dispatch:
+    name: "tell the store"
+    runs-on: ubuntu-latest
+    # Repos in an org without a session-memory store show this SKIPPED
+    # rather than green. The variable carries owner/repo, set once at
+    # org level — deliberately not baked in at stamping time, so the
+    # rendered file names no org's store and a store move is one
+    # variable edit, not a re-stamp of every repo.
+    if: ${{ vars.SESSION_MEMORY_REPO != '' }}
+    steps:
+      - name: Name the store's repository
+        id: store
+        env:
+          STORE: ${{ vars.SESSION_MEMORY_REPO }}
+        run: |
+          set -euo pipefail
+          echo "name=${STORE##*/}" >> "$GITHUB_OUTPUT"
+
+      # The same app that runs releases and template updates, because its
+      # key already reaches every stamped repo and — unlike a fine-grained
+      # PAT — never expires, so no yearly morning where every sender in
+      # the org goes red at once. repository_dispatch is outside the
+      # repo-scoped GITHUB_TOKEN's reach; the minted token is scoped to
+      # the store repo alone.
+      - name: Mint a token that can reach the store
+        id: app-token
+        uses: actions/create-github-app-token@fee1f7d63c2ff003460e3d139729b119787bc349 # v2.2.2
+        with:
+          app-id: ${{ vars.RELEASE_BOT_CLIENT_ID }}
+          private-key: ${{ secrets.RELEASE_BOT_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: ${{ steps.store.outputs.name }}
+
+      - env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          STORE: ${{ vars.SESSION_MEMORY_REPO }}
+          TICKET: ${{ github.repository }}#${{ github.event.issue.number }}
+        run: |
+          set -euo pipefail
+          # The payload carries only the ticket's name. The store
+          # re-reads the ticket's state itself, so a stale or replayed
+          # dispatch can cause a wasted lookup, never a wrong event.
+          gh api "repos/${STORE}/dispatches" \
+            -f event_type=ticket-closed \
+            -f "client_payload[ticket]=${TICKET}"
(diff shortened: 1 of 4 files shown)


.github/workflows/ticket-closed.yml
```diff
@@ -0,0 +1,47 @@
+name: Report ticket close to the ledger
+
+# The session-memory store closes a thread when the thread's own ticket
+# closes. The store cannot hear this repository's events — a workflow
+# only receives events from the repository it sits in — so every repo
+# that mints tickets reports its own closes. Event-driven end to end:
+# the alternative was the store polling every ticket on a clock, ruled
+# out for its latency (decision-memory:
+# close-loop-trigger-poll-overruled).
```
pando-genet: Remove the text about the alternative, it is irrelevant

.github/workflows/ticket-closed.yml
```diff
@@ -0,0 +1,47 @@
...
+    runs-on: ubuntu-latest
+    # Repos in an org without a session-memory store show this SKIPPED
+    # rather than green. The variable carries owner/repo, set once at
+    # org level — deliberately not baked in at stamping time, so the
+    # rendered file names no org's store and a store move is one
+    # variable edit, not a re-stamp of every repo.
+    if: ${{ vars.SESSION_MEMORY_REPO != '' }}
+    steps:
+      - env:
+          # A fine-grained PAT with Contents read/write on the store —
```
pando-genet: Isn't this superseded by the release bot app?

template/{% if agentic_forge == 'github' %}.github{% endif %}/workflows/ticket-closed.yml
pando-genet, line 1: Why do we need this file twice m

.github/workflows/ticket-closed.yml
pando-genet: Removed in `ae3ac00`, from both copies. The header now states only what the workflow does; the design history stays in decision-memory.

---

.github/workflows/ticket-closed.yml
pando-genet: Yes — if this is the fine-grained-PAT text, it is superseded and already gone: `c6a8af1` replaced the PAT with the release-bot mint, and the file at the PR head carries no PAT reference (the old block only appears when browsing the first commit alone; the Files-changed view shows the end state).

If it points at the `Name the store's repository` step instead: that one is not superseded — it feeds the mint's `repositories:` input, which takes a bare repo name while `SESSION_MEMORY_REPO` carries `owner/repo`, and it is what keeps the minted token scoped to the store repo alone instead of everything the app can reach.

---

template/{% if agentic_forge == 'github' %}.github{% endif %}/workflows/ticket-closed.yml
pando-genet: Because this repo is the template AND its own consumer. The copy under `template/` is the source that ships to stamped repos; the copy at the root is this repo applying its own template to itself — same as `add-to-project.yml` and `labels.yml`, which exist twice for the same reason. `tests/test_self_application.py` enforces that every root file with a `template/` counterpart is byte-identical to the render, so the two cannot drift: a template file with no adopted root copy fails the gate as `missing`, forcing exactly the decision made here (adopt it, or list it under `DELIBERATE_DIVERGENCE` with a reason). Adopting is right for this file — this repo mints tickets that ledger threads reference, so its closes need reporting like any other repo's.

---

pando-genet:
Restructured onto the documented route (`docs/conventions.md` §Template-First Changes). The branch is now two commits:

- `d9be9a3` **feat** — `template/` and tests only, no root file
- `b275623` **chore: reapply template to self** — the root file alone, and it is the *render output*: produced by `uv run copier copy --trust --defaults --skip-tasks --vcs-ref HEAD …` into a temp dir and adopted verbatim, not hand-copied

I had violated both halves of that rule — `cp` from the template, consolidated into the feature commit. Reworked rather than patched on top, so the history shows the route rather than a correction to it.

Verified the split is load-bearing, not ceremonial: checked out the intermediate commit and ran the gate, which fails exactly as designed —

```
AssertionError: Rendered files absent from the repo root:
['.github/workflows/ticket-closed.yml'].
Adopt them, or add them to DELIBERATE_DIVERGENCE with a reason.
```

So `d9be9a3` is a deliberate red on the self-application gate, resolved by `b275623`. CI evaluates at PR HEAD, where the full suite is green: 206 passed, 4 skipped, lint clean.

---

PR #132, opened 2026-08-04
Stamp session-memory stores from a subtemplate

pando-genet:
Refs pandoscope/session-memory#2.

A session store was the only memory store with no subtemplate, so it could not be stamped and its files were maintained by hand. This adds `agentic_subtemplate=session-memory`, modelled on the evidence store.

## What it ships

`AGENTS.md`, `CLAUDE.md`, `README.md`, `.gitignore`, `template-update.yml`, `ticket-closed.yml` — and nothing else. **Deliberately thinner than the other stores**: the recorder that writes a session store lives with the `thread-ledger` skill, so the schema keeps exactly one home and this subtemplate vendors no copy of it. That is the difference from `decision-memory` and `evidence-memory`, which vendor their writers and validators because those have no other home.

## The bigger find: no stamped store had the sender

Chasing this turned up a wider hole than the missing subtemplate. `ticket-closed.yml` shipped only in `template/`, so **neither `decision-memory` nor `evidence-memory` would ever have carried it** — and both mint tickets that ledger threads reference (`decision-memory#5` is live right now). A store without the sender is a hole in the close-loop, not a repo that happens not to need one.

So the sender now reaches all three stores, pinned byte-identical across all four copies by `test_shared_workflows_match_the_ones_consumers_get` — the same managed-duplication treatment `template-update.yml` already gets, because copier renders one `_subdirectory` at a time and a store cannot import the consumer copy. `test_every_store_reports_its_ticket_closes` fails if a future store subtemplate forgets it.

## The README is seeded, not vendored

`_skip_if_exists` for this subtemplate only. A store's README says which siblings it sits beside and what its threads are about; vendoring it would replace that with the generic seed on every `copier update`. `test_the_readme_is_seeded_not_vendored` pins that by recopying over an edited README and asserting the edit survives.

Note the shipped README and AGENTS.md name **no organisation** — `test_the_store_names_no_organisation` enforces it. The sender resolves its destination from `vars.SESSION_MEMORY_REPO` at run time, so a stamped file naming one org would be wrong for every other and stale for this one.

## Testing

`tests/test_session_memory_subtemplate.py`, 8 cases: exact file manifest, minimal answers file, the seeded-README contract, the no-org rule, store rules absent from consumer renders, and the two pinning tests above.

The two existing store manifests failed until they declared the new sender — the gate doing exactly what it is for, so both were updated rather than loosened.

Full suite: **214 passed, 4 skipped**, lint clean. No self-application commit: the subtemplate is a source directory with no counterpart at this repo's root, and the gate confirms it.

## Follow-up, not in here

Stamping the actual `pandoscope/session-memory` repo with this — a separate PR in that repo once this lands.

---

### DIFF
--- a/copier.yml
+++ b/copier.yml
@@ -155,11 +155,6 @@ _skip_if_exists:
   # budget a human tuned.
   - "{% if agentic_subtemplate == 'decision-memory' %}preferences.md{% endif %}"
   - "{% if agentic_subtemplate == 'decision-memory' %}store.config.json{% endif %}"
-  # The session store's README describes THAT store — which sibling
-  # stores it sits beside, what its threads are about. Vendoring it
-  # would overwrite that with the generic seed on every update, so it
-  # is seeded once and owned by the store thereafter.
-  - "{% if agentic_subtemplate == 'session-memory' %}README.md{% endif %}"
 _jinja_extensions:
   - copier_template_extensions.TemplateExtensionLoader
   - extensions/agentic.py:AgenticExtension
--- a/copier.yml
+++ b/copier.yml
@@ -9,12 +9,17 @@ agentic_subtemplate:
     vendored store docs, and a seeded preference set;
     `evidence-memory` stamps an evidence STORE — the capture writer, the
     CI guards (validator + guard runner + workflow), and the vendored
-    store docs. With either store subtemplate, all project-scaffold
-    questions are skipped, keeping the store's answers file minimal.
+    store docs;
+    `session-memory` stamps a session STORE — the store's agent rules
+    and the workflow that reports its ticket closes to the ledger; its
+    recorder ships with the thread-ledger skill, not from here. With
+    any store subtemplate, all project-scaffold questions are skipped,
+    keeping the store's answers file minimal.
   choices:
     - template
     - decision-memory
     - evidence-memory
+    - session-memory
   default: template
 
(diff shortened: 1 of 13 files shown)


pando-genet:
The README is now **vendored, not seeded** (`4ff6bbe`) — the principal's call, and the better one: the store holds data, not the source of its own documentation.

- `_skip_if_exists` entry dropped, so the README travels like `AGENTS.md`.
- `test_the_readme_is_seeded_not_vendored` flipped to `test_the_readme_is_vendored`: an edit made in the store is now *replaced* on recopy, and the test asserts that rather than its opposite.
- The vendored text folds in what the first store's README had earned — the reporting rule for unreadable tickets, the transcript-export contract, the note that a vector index has to beat a measured miss rate against a hand-read sample — generalized so it names no organisation, which `test_the_store_names_no_organisation` enforces.

What seeding would have cost, stated plainly since I argued for it an hour ago: N stores forking N descriptions of one contract, and the store that edited its copy being the one least likely to notice the schema had moved on.

One consequence to expect when the store is stamped: its current README links to its sibling stores by URL, and the vendored text names them without links, because a vendored file cannot carry one org's URLs. If those links are worth keeping, they belong somewhere org-specific rather than in a file every store shares.

Suite still 214 passed, 4 skipped; lint clean.

---

Issue #137, opened 2026-08-04
One uniform CI gate: an aggregate check every repo exposes under the same name, plus ticket and fixup gates


pando-ramet:
Blocks the last three items of the principal's confirmed merge policy (pandoscope/meta#57 carries the rest as canonical rulesets):

> only merge after successful CI, history unsquashed but **free of fixups**, merge commit references the PR, and **that PR must have a ticket assigned**

Signatures, PR-only, merge-commits-only, no-force-push and review-thread resolution are all encoded and applied. These three are not, and none of them can be a ruleset rule alone.

## Why a ruleset cannot express "green CI" today

`required_status_checks` names **exact check contexts**, and every repo's differ:

| Repo | Checks |
| --- | --- |
| skills | `bats + shellcheck`, `prek (all hooks, all files)`, `commitlint (PR commits)` |
| disambiguate | `test (3.13, ubuntu-latest)`, `lint`, `Lint Commit Messages`, `prek …` |
| ghx | `BATS test suite`, `prek …` |
| agentic-engineering-template | `Template: 3.12 - ubuntu-latest` ×4, `Lint …`, `Conflict detection …` |

Declaring those per repo shatters one canonical body into N variants that drift the moment a job is renamed — the exact failure the ruleset sync exists to prevent. Worse, a **renamed job silently stops being required**: the ruleset keeps waiting for a context nobody produces, and GitHub treats an absent required check as pending, not failed, so the gate quietly becomes unmergeable-forever or (if removed) no gate at all.

## The fix: one aggregate job, template-shipped

Ship a `ci-ok` job in the consumer template that `needs:` every other job in the repo's workflows and fails if any dependency did not succeed:

```yaml
  ci-ok:
    name: "ci-ok"
    if: always()
    needs: [test, lint, commitlint, prek]   # rendered per repo's real jobs
    runs-on: ubuntu-latest
    steps:
      - name: Every required job succeeded
        run: |
          set -euo pipefail
          # `needs.*.result` covers cancelled and skipped too: a job that
          # did not run is not a job that passed.
          echo '${{ toJSON(needs) }}' | jq -e 'all(.[].result == "success")'
```

Then `required_status_checks` names exactly `ci-ok`, everywhere, forever — and adding or renaming a job changes the `needs:` list, not the ruleset. **`if: always()` is load-bearing**: without it the aggregate is skipped when a dependency fails, and a skipped required check reads as pending, so the PR hangs instead of reporting red.

[...]

PR #144, opened 2026-08-06
feat(template): one uniform CI gate — ci-ok, ticket references, review answers

### LINKED ISSUE #143: review-answers gate: every reviewer thread answered with a verified commit link or an explicit no-commit marker
Principal's ask (2026-08-06): every review comment by the principal on an agent-authored PR is answered by the agent with a **proper GitHub commit link** (full URL, not a bare hash) or an **explicit statement that no commit was required** — and the answering is checked mechanically.

## Why this is a CI gate, not a heartbeat check

The heartbeat's contract is observed local state and it runs in managed sessions where GitHub CLIs are sabotaged and hooks hold no API token — review threads are server-side, and a check that cannot enumerate them would guess. The enforcement moment for review conversations is merge time, which is exactly where the #137 `ci-ok` vehicle already stands.

## The job

`review-answers`, in the template's aggregated workflow (rides `ci-ok`, #137), triggered on `pull_request` + review/review-comment events:

- Enumerate the PR's review threads. For each thread **started by a human reviewer** (not the PR author, not a bot): require a **later reply by the PR author** containing one of
  - a full commit URL `https://github.com/<owner>/<repo>/commit/<sha>` where `<sha>` is reachable on the PR head branch — verified via the API, so a pasted-but-wrong link fails; a bare hash fails by design (not clickable, not verifiable at a glance);
  - the canonical no-commit marker: a reply line starting `No commit:` followed by the reason. Exact prefix, greppable — a loose "no commit was needed" match would also be satisfied by "no commit yet".
- Threads resolved without any qualifying reply still fail — resolving is not answering.
- Green when no human review threads exist.

## Heartbeat companion (best-effort, separate slice)

Review comments arrive in the session transcript as webhook events and replies are tool calls in the same transcript — both local state. A turn-time nudge can hold a turn that received review comments to having answered in the required form before it ends. Heuristic by nature (out-of-session replies are invisible), so it nudges; the CI job gates. Slice separately in skills once the gate exists.

## Predicted decisions (overrulable)

- Marker phrase exactly `No commit:` (prefix, case-sensitive like the ALL-CAPS keyword list).
- "Human reviewer" = any review-thread author who is not the PR author and not a bot — not hardcoded to one username; the principal/agent pair is the org's current shape, not the rule.
- Commit-link verification = sha reachable from the PR head (compare API), so links to commits on other branches fail.

## Related

- #137 — the `ci-ok` aggregate this rides; same workflow file, same rollout.
- #134 / #142 — the update workflow the template ships beside it.

pando-ramet:
ADVANCES #137. CLOSES #143.

## What

`ci-ok.yml` — one workflow, three jobs, one testable script (`scripts/ci/check_gate.py`, stdlib+yaml, pure decision functions behind a thin fetch layer):

- **`ci-ok`** — the single context `protect-main` will require, everywhere, forever. Discovers every `pull_request`-triggered workflow on the head SHA **at runtime** (trigger `types` honored: a workflow that cannot produce a run for this SHA — e.g. `types: [opened]` — is not awaited), polls until all their jobs complete, and fails on any non-success conclusion. Skipped is not passed. Its own run is judged sibling-by-sibling, so it never waits for itself. Adding or renaming a job changes nothing here — the within-workflow *and* cross-workflow drift classes close together, which supersedes the separate static meta-check the rulings anticipated (runtime discovery covers both once expectations come from the tree).
- **`ticket`** — PR bodies reference tickets in exact ALL CAPS from the central `.github/reference-keywords.json`: `CLOSES`/`FIXES` (closing) and `ADVANCES` (non-closing, deliberately not a native keyword). Any of GitHub's nine native closing keywords in any other casing **fails** the PR — the forge would act on `resolves #n` whether or not the gate recognized it. Every ticket number in a `claude/<n>[-<n>]-…` branch must appear as a canonical reference in the body. The only escape: the `automated` label on a bot-authored PR — an agent doing manual work can hold neither half. Reads only the event payload; label/body edits re-run just this workflow (widened trigger types), never the test matrix.
- **`review-answers`** (#143) — every review thread opened by a human needs a later PR-author reply carrying a **full commit URL verified against the PR's own commits** (bare hashes and pasted-but-wrong links fail; the URL base comes from `GITHUB_SERVER_URL`, not a hardcoded vendor) or an explicit `No commit: <why>` line. Resolving a thread is not answering it — the check never reads resolution state.

Also:

- **commitlint for every consumer** (`lint.yml.jinja`): full conventional commits over the PR range — the type enum is the allow-list, so `fixup!`/`squash!`/`wip` fail by grammar, no deny rules needed.
- **Store subtemplates** (decision-/evidence-memory) carry the gate **without the ticket job** — stores are not ticketed (ruling). session-memory stays excepted from protect-main entirely.
- **AGENTS.md** teaches the same rules the gate enforces, from the same central file.
- Self-restamped in its own commit per convention.

## What this PR does NOT do (the rollout, in order)

1. Release + fan-out delivers the gate to consumers (this also live-proves #142's supersede path).
2. meta and remaining non-consumers become template consumers (ruling D3) — separate PRs.
3. Only after **every** repo has reported a green `ci-ok` on main: the meta PR adding `required_status_checks: ["ci-ok"]` to `protect-main.json`. Flipping earlier blocks every open PR on a context that does not exist.

The heartbeat-side best-effort review-answer nudge is a separate skills slice, per #143.

## Tests

`tests/test_ci_gate.py`, 23 cases over the pure functions: canonical/cross-repo/ADVANCES pass; lowercase and unlisted native keywords fail even beside a canonical ref; wrong-case of our own keywords fails; branch↔body number matching; bot-only escape; verified-URL vs bare-hash vs wrong-sha answers; `No commit:` exact vs loose wording; own/bot threads ungated; thread grouping; workflow discovery incl. YAML-1.1 `on`; red/skipped/pending/green aggregate verdicts; own-run sibling judgment; byte pins template-first across the root and store copies. Full suite (minus e2e) 237 pass; prek incl. actionlint clean.

**Honest gap:** the aggregate's poll loop against the live Actions API is exercised only by the pure-verdict tests plus this PR's own runs — the first genuinely red dependency in the wild is worth watching once.

---


template/AGENTS.md.jinja
```diff
@@ -142,6 +142,18 @@ Code-specific skills:
 - Never push to `main`
 - Create PR immediately on branch creation
 - Commits: conventional commits
+- **Ticket references in PR bodies are ALL CAPS, from the central list**
+  (`.github/reference-keywords.json`, enforced by the `ticket` job):
+  `CLOSES #n` / `FIXES #n` close on merge, `ADVANCES #n` references a
```
pando-genet: Inject the allowed terms from a central location to avoid drift

template/AGENTS.md.jinja
```diff
@@ -142,6 +142,18 @@ Code-specific skills:
...
+  multi-PR ticket without closing it. Any native closing keyword in any
+  other casing fails the gate — the forge would act on `resolves #n`
+  whether or not the gate recognized it. Every ticket number in the
+  branch name must appear as a canonical reference in the body.
+- **Answer every review comment with a full commit URL or `No commit: <why>`**
```
pando-genet: Inject the pattern from a central location here and in code

pando-genet (changes requested):

pando-ramet:
Both review comments addressed in [`e24c0af`](https://github.com/pandoscope/agentic-engineering-template/commit/e24c0af5c7b12e8db93d94c465eb5b3d52d56a0d) (template) + [`5be55ee`](https://github.com/pandoscope/agentic-engineering-template/commit/5be55ee76753b76d16e334dcc1a8f3b1839fa1ea) (restamp):

- **Allowed terms injected from the central location:** `AGENTS.md.jinja` now renders its reference rules — the keyword lists, the native deny list, the branch pattern, the `No commit:` marker — from `reference-keywords.json` via a `reference_keywords()` jinja global in the copier extension. The docs and the gate read the same bytes; nothing left to drift.
- **The pattern injected here and in code:** `reference-keywords.json` gains `branch_pattern` and `no_commit_marker`; `check_gate.py` reads every rule from the file (a repo without it — the store variants carry no ticket gate — falls back to built-ins), and the rendered AGENTS.md carries the same values.

Tests gain the config-driven cases (custom branch pattern honored; custom marker honored); 238 pass, prek clean.

---

template/scripts/ci/check_gate.py
```diff
@@ -142,13 +141,14 @@ def thread_of(comments):
     return list(threads.values())
 
 
-def review_violations(threads, pr_author, pr_shas, base_url):
+def review_violations(threads, pr_author, pr_shas, base_url, marker="No commit:"):
```
pando-genet: Don't set a default, require to be always explicitly passed from central location only!

pando-genet (changes requested):
Remember to put restamp commit last. Address comment

pando-ramet:
Both review points addressed; branch rebuilt with the restamp commit last:

- **No defaults, central location only:** [`2494579`](https://github.com/pandoscope/agentic-engineering-template/commit/2494579279c261b4822ce154dcd4c8cc4bdaf663) — `check_gate.py` requires `reference-keywords.json` for every rule value (`branch_pattern`, `no_commit_marker`, the keyword lists): missing file or key fails loudly, no built-in fallback anywhere, and the workflows pass the path explicitly. That made the store variants carry the same keyword file (byte-pinned like the script), which is cleaner than the fallback they were leaning on. New test pins the no-defaults behaviour (`test_no_built_in_defaults_every_rule_is_explicit`).
- **Restamp last:** history is now exactly two commits — the template authorship, then [`285503d`](https://github.com/pandoscope/agentic-engineering-template/commit/285503dbe5823afd0ce7b950b9d0bbc0f2659b39) `chore: restamp the repo from its own template` as the final commit.

239 tests pass, prek clean.

---

pando-ramet:
Addressed in [`bd82deb`](https://github.com/pandoscope/agentic-engineering-template/commit/bd82deb047fd891e2ae6e84cc4a395e5bfba6cf0) (restamp [`87375e3`](https://github.com/pandoscope/agentic-engineering-template/commit/87375e37e06fed99bf0d01e8dfb8cdc6234dc1a1) last): workflow comments and the script docstring now *refer* to `reference-keywords.json`'s keys (`no_commit_marker`, the allow-list, the branch pattern) instead of quoting their values — the marker and keyword enumerations appear verbatim nowhere outside the central file. The root AGENTS.md already rendered from it; this round the rendered copies were adopted from a fresh copier self-render. 25 gate tests, prek clean.

---

pando-ramet:
Right — they're templates, so the values themselves are injected where readers look. In [`bca9ac3`](https://github.com/pandoscope/agentic-engineering-template/commit/bca9ac382a6c3b9ab594452aee2f6a5d6a7137f0) (restamp [`5a30ec1`](https://github.com/pandoscope/agentic-engineering-template/commit/5a30ec19795fad661987d7abe08972aaf5ecd899) last): `ci-ok.yml` became `ci-ok.yml.jinja` — the workflow comments interpolate the actual marker and the actual ALL-CAPS list from `reference_keywords()` at render time, so a rendered repo's workflow reads plainly with zero indirection while the single source stays `reference-keywords.json`. Store variants same. The rendered root copy shows the literal marker in its comment; the pin test asserts exactly that (value present, no jinja residue). AGENTS.md was already injected this way. 239 tests, prek clean.

---

pando-genet (approved):

PR #146, opened 2026-08-07
fix(template): the ticket gate's escape reaches the update PR, and commitlint finds its preset

pando-ramet:
## What

Two defects the v3.14.0 fan-out surfaced the moment the gate met its own delivery vehicle — three of six consumer update PRs (skills, disambiguate, ghx) went red on the gate's own checks, with zero copier conflicts:

1. **The ticket gate's designed escape was unreachable.** `check_gate.py` already exempts a bot-authored PR carrying the `automated` label — but nothing ever applied that label, so every template-update PR failed its own template's gate.
2. **commitlint died with MODULE_NOT_FOUND** in every repo without a `node_modules` of its own: it resolves the config's `extends` preset from the *repo directory*, not npx's cache, so the cache-only `-p` install never worked outside repos that happen to install node deps.

ADVANCES pandoscope/agentic-engineering-template#137

## How

- `template-update.yml` (template + the three byte-identical store copies): `gh label create automated || true` before `gh pr create --label automated` — bootstrapped inline because this very PR can be the one delivering `labels.toml`'s entry to the repo.
- `labels.toml` (template + AET's own): the `automated` entry, so the label sync never deletes what the workflow creates. The evidence store's minimal label set stays untouched — its gate runs no ticket job, so the label would have no consumer there (its own test enforces exactly that).
- `lint.yml.jinja`: `npm install --no-save` the commitlint packages into the checkout, then `npx --no-install commitlint` — the preset now resolves from where commitlint actually looks.

## Proof

- `test_generate_project.py` pins both mechanics in the rendered output: the label bootstrap + `--label automated` in the update workflow, and the local-install commitlint invocation.
- Full suite: 241 passed (e2e deselected locally, runs in CI).
- Live proof comes with the release: the v3.14.1 update PRs themselves must arrive labeled and pass commitlint — the gate's own delivery vehicle is the test case.

---

### DIFF
--- a/template/{% if agentic_forge == 'github' %}.github{% endif %}/workflows/template-update.yml
+++ b/template/{% if agentic_forge == 'github' %}.github{% endif %}/workflows/template-update.yml
@@ -116,14 +116,13 @@ jobs:
           token: ${{ steps.app-token.outputs.token }}
           fetch-depth: 0
 
-      # An open update PR used to stand the run down wholesale — a
-      # deadlock, not a guard (#134): a stale PR nobody merged blocked
-      # every subsequent release forever, and the repo most in need of
-      # an update was the one guaranteed not to get one. An update PR
-      # for the CURRENT release is genuine dedupe; one for an older
-      # release is superseded — closed with a comment — and the run
-      # proceeds to open the current one.
-      - name: Skip on a current update PR, supersede a stale one
+      # An open update PR for the CURRENT release is genuine dedupe. A
+      # stale one is JOINED (#137): its branch is checked out and the
+      # new release lands as a commit ON it — the PR, its discussion and
+      # the conflict resolutions already on the branch all survive.
+      # Closing it would throw that branch work away; skipping wholesale
+      # would let a PR nobody merged block every later release (#134).
+      - name: Skip on a current update PR, join a stale one
         id: dedupe
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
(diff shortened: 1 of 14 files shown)


pando-genet (changes requested):
Stamp template in separate commit

pando-ramet:
Restructured as requested: the template change (`template/` + its generate-project pins) is now `fcbc189`, the restamp of this repo's own `.github` and the three store copies is a separate `chore: restamp the repo from its own template` (`5e97f8e`), and the e2e expected-tree fix stays its own commit (`8d5eebc`). The tree is byte-identical to the reviewed head — `git diff` between the two is empty; only the commit boundaries moved.

---

pando-ramet:
Added per your direction: a stale update PR is now **joined, never closed** (`d7a0e9a` + its restamp `56e91d9`). The dedupe checks the stale PR's branch out, `copier update` runs on top of it, the new release lands as a Verified `createCommitOnBranch` commit on the existing branch, and the PR's title, body and label refresh in place — so its discussion and any conflict resolutions on the branch survive the new release. A PR for the current release still skips; a repo with no open update PR keeps the create path. Recorded as `20260807T045541Z-stale-update-pr-joined-not-closed`.

---

template/{% if agentic_forge == 'github' %}.github{% endif %}/workflows/template-update.yml
```diff
@@ -117,13 +117,14 @@ jobs:
...
 
       # An open update PR used to stand the run down wholesale — a
-      # deadlock, not a guard (#134): a stale PR nobody merged blocked
-      # every subsequent release forever, and the repo most in need of
-      # an update was the one guaranteed not to get one. An update PR
-      # for the CURRENT release is genuine dedupe; one for an older
-      # release is superseded — closed with a comment — and the run
-      # proceeds to open the current one.
-      - name: Skip on a current update PR, supersede a stale one
+      # deadlock, not a guard (#134) — and then a stale one was
```
pando-genet: It doesn't matter what once was, only current situation is relevant

pando-genet (changes requested):

template/{% if agentic_forge == 'github' %}.github{% endif %}/workflows/template-update.yml
pando-ramet: Rewritten in https://github.com/pandoscope/agentic-engineering-template/commit/ddb26c10a6951e8fbc9c7db19fbfa14e679a7f48 — both comment blocks now state only the current behavior and its rationale (join on stale, dedupe on current, and why the two alternatives are unacceptable), with no narration of what the workflow once did. The in-place-refresh comment likewise describes the invariant ("one PR carries the update conversation however many releases pass through it") rather than the history.

---

pando-genet (approved):

Issue #129, opened 2026-08-04
Tracker content: a commit named in a comment must be clickable


pando-genet:
Reported by the principal on pandoscope/meta#45: two review replies cited the commits that fixed the review, and neither could be clicked.

## The trap

The tracker autolinks a **bare** SHA in the same repo. Backticks suppress it — and code-formatting an identifier is the reflex for every other identifier in a comment, so the habit is precisely what breaks the link. It also looks *more* correct while doing so. A cross-repo SHA never autolinks at all, bare or not.

That puts it in the same class as the two rules already in `### Tracker Content Formatting`: content that renders correctly in a committed `.md` file and silently degrades when posted to a tracker.

## What to write

In `template/AGENTS.md.jinja`, under `### Tracker Content Formatting`:

- A commit named in tracker-posted content is written as a link — `[«short-sha»](«repo-url»/commit/«full-sha»)`, or `«owner»/«repo»@«sha»` for another repo.
- Why backticks are the trap, so the rule survives someone deciding it looks untidy.
- The general case: anything the reader will want to open — a file (blob at a ref), a workflow run, a cross-repo PR or issue — gets a link. A reference that has to be copy-pasted into a search box will not be followed.

## Acceptance

- [ ] the rule is in `template/AGENTS.md.jinja`, in the tracker section
- [ ] the repo's own `AGENTS.md` carries it via self-application, as its own `chore: reapply template to self` commit — not a hand-edit
- [ ] `tests/test_self_application.py` green

## Not in scope

Sweeping existing tickets. That section's own repair rule says a violation in an existing ticket is raised with the user rather than rewritten unprompted.

## Related

- pandoscope/meta#45 — where it was reported
- **Failures Become Rules** in `AGENTS.md`, which names the template as the preferred home so every generated repo inherits the fix

PR #147, opened 2026-08-07
feat(dev): the self-application route is enforced, not conventional

### LINKED ISSUE #130: Self-application: guard the route, not just the result — a commit may not edit a templated root file and its template source together
Found while auditing why the self-application gate did not catch a violation of the self-application *route* on #128.

## What happened

#129 was first built as one commit that hand-edited **both** `template/AGENTS.md.jinja` and the root `AGENTS.md`. The principal caught it by reading the diff; the rule it broke is `docs/conventions.md` § Template-First Changes — root files with a `template/` counterpart are render *output*, never hand-edited.

CI was green throughout, and correctly so.

## Why the existing gate cannot catch this

`tests/test_self_application.py` asserts `root == render`. That is a statement about **state**. The rule that was broken is about **provenance**. When the hand-edit happens to be byte-identical to what the render would have produced, the state is valid and there is nothing to detect.

Measured, not assumed:

- The wrong commit `b19d0cd` and the correctly-rebuilt pair `6af1cd0` + [eaf9b84](https://github.com/pandoscope/agentic-engineering-template/commit/eaf9b849e66ff66e810ec82b6b15ef6150c65106) produce the **same tree**, `7c0469255301a16bec277ed75a4fe60c0b515ec5`. Both CI runs ([151](https://github.com/pandoscope/agentic-engineering-template/actions/runs/30886947744), [152](https://github.com/pandoscope/agentic-engineering-template/actions/runs/30888406238)) are green. The gate had nothing to say.
- The gate does fire when state is actually wrong: reverting root `AGENTS.md` to its pre-change content while leaving the template edited → `FAILED`. Rendering at `6af1cd0` (template edited, root not yet stamped) → `FAILED`.

So the gate works exactly as designed. It is the wrong instrument for this failure, and no content comparison can be the right one.

## Proposal

A `prek` hook: **reject a commit whose diff touches both `template/«path».jinja` and the root `«path»`.**

The two-commit route makes this trivially safe — under the documented route, a template commit touches only `template/**` and the `chore: reapply template to self` commit touches only root. A commit touching both is, by construction, either a hand-edit or a mixed commit, and both are already forbidden.

Exempt the paths in `DELIBERATE_DIVERGENCE` (`tests/test_self_application.py`): those legitimately diverge from the render and may need editing alongside their template counterpart.

This catches the provenance violation at the moment it is made, in the one place where the information still exists — the commit boundary. After the fact it is unrecoverable, which is why review was the only thing that caught it.

## Acceptance

- [ ] Hook rejects a commit touching `template/X.jinja` and root `X` together, naming both paths and pointing at `docs/conventions.md` § Template-First Changes
- [ ] `DELIBERATE_DIVERGENCE` paths exempt, sourced from the test rather than duplicated
- [ ] A test proves the hook fires on the #129 shape and stays quiet on both legs of the correct two-commit route
- [ ] `docs/conventions.md` notes that the route is now enforced, not only conventional
- [ ] `prek run --all-files` clean; `uv run pytest` green

## Not in scope

Making CI *perform* the self-application (auto-render and commit the root). That would remove the reviewable `chore: reapply template to self` commit, which is the artifact that makes the stamp auditable.

## Related

- #114 — rules that self-application is verified at HEAD, not per commit. Confirmed still accurate here: `6af1cd0` genuinely fails the gate and CI never sees it, because Actions triggers on the branch head only. That ticket's documentation work is still open and this hook does not replace it — #114 is about *when* the state is checked, this is about *how* the state was reached.
- #129 / #128 — where the route was violated
- `docs/conventions.md` § Template-First Changes — the rule this enforces

pando-ramet:
## What

A root file with a `template/` counterpart is render output, so a commit edits the source or the stamp, never both. That rule was conventional; it is now checked.

CLOSES pandoscope/agentic-engineering-template#130

## Why state cannot catch it

`test_self_application` compares root against the render — a statement about **state**. The route is about **provenance**, and no content comparison can see it: a hand-edit that happens to produce exactly what the render would have leaves a valid tree. The ticket measured this (`b19d0cd` and the correct pair produce the same tree, both green). The commit is the only place the information still exists, so that is where this looks.

## How

`scripts/dev/self_application.py` — one pure rule (`stamp_of`, `route_violations`) with two entry points:

- **The `self-application-route` prek hook** at commit time. It reads the staged diff itself rather than the filenames prek passes, which is what keeps `prek run --all-files` quiet: with nothing staged there is no commit to judge.
- **A `Template-first route (per commit)` CI job** over every commit a PR adds. Hooks run only in clones that installed them — both violations that motivated this ticket were caught by human review, and a third slipped through on !146 in a container whose clone has no hook installed. This is the arm that actually bites.

Exempt by construction, because no route governs them: paths that deliberately diverge, and paths copier seeds once (`_skip_if_exists` — after seeding, the root file is this repo's own document; editing it beside its template origin is legitimate).

**One deviation from the acceptance wording**, flagged for your call: the criterion says `DELIBERATE_DIVERGENCE` should be "sourced from the test". Importing `tests/test_self_application.py` inside a commit hook would drag `copier` and `yaml` into every commit, so the direction is inverted instead — the list moves to `scripts/dev/self_application.py` and the test imports it from there. Single source preserved, which is the criterion's intent; `docs/conventions.md` now points at the new home. Same reasoning for `_skip_if_exists`: read without a YAML parser so a commit hook never fails for want of a dependency, and pinned in tests against a real parse of the same block.

## Proof

- 14 route tests: the #129 shape fires; both legs of the documented two-commit route stay quiet; a deliberate divergence and a seeded file are exempt; every violation is reported, not just the first; the divergence list has one home (equal contents **and** no second literal); the no-YAML read of `_skip_if_exists` equals the real parse.
- Exercised against real git history in a scratch repo: staged violation → exit 1 naming both paths, correct two-commit route → exit 0 on each leg and across both, `--range` including the violation → exit 1 naming the commit.
- `prek run --all-files` clean, `uv run pytest` green (254 + 9 e2e).
- The guard passes on its own commit — this change touches no `template/` source.

## Not in scope

Making CI *perform* the self-application, per the ticket. Also left alone: the store subtemplates' byte-pinned copies (`decision-memory/scripts/ci/check_gate.py` and friends) are a second stamp relationship, but same-named files there are not all copies — some are the subtemplate's own — so extending the rule to them needs a declared list first. Worth a follow-up ticket.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,30 @@ jobs:
       - run: prek run --all-files --show-diff-on-failure
         shell: bash
 
+  template-route:
+    # The route guard where it actually bites. prek hooks run only in
+    # clones that installed them, and a fresh container or a CI-only
+    # contributor has not — the two violations that motivated #130 were
+    # both caught by human review, not by a hook. Every commit the PR
+    # adds is judged on its own diff, so the boundary information is
+    # still there when this runs.
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    name: "Template-first route (per commit)"
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
+        with:
+          python-version: "3.13"
+      - name: No commit edits a template source and its own stamp
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: python3 scripts/dev/self_application.py --range "$BASE_SHA..$HEAD_SHA"
+        shell: bash
+
   conflict-detection:
     # Enforcement arm of the file-ownership policy: copier update merges and
     # writes .rej files but still exits 0. Generate a fixture, run the update
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,6 +8,20 @@ ci:
   autoupdate_commit_msg: "chore(pre-commit.ci): pre-commit autoupdate"
 
 repos:
+  - repo: local
+    hooks:
+      # The self-application route, checked where the information still
+      # exists (#130). A root file with a `template/` counterpart is
+      # render output, so a commit edits the source or the stamp, never
+      # both. Reads the staged diff itself rather than the filenames
+      # prek passes, which is what keeps `--all-files` runs quiet: with
+      # nothing staged there is no commit to judge.
+      - id: self-application-route
+        name: "template-first route (source and stamp never in one commit)"
+        entry: python3 scripts/dev/self_application.py
+        language: system
+        pass_filenames: false
+        always_run: true
   - repo: https://github.com/alessandrojcm/commitlint-pre-commit-hook
     rev: v9.24.0
     hooks:
--- a/docs/conventions.md
+++ b/docs/conventions.md
@@ -37,10 +37,18 @@ For any change touching a templated file, always follow this route:
 This doubles as a proving ground: the self-applied root files are the rendered
 template output, reviewed in the same PR as the template change.
 
+The route is enforced, not merely conventional: a commit that edits a template
+source and the root file it stamps is rejected, by the `self-application-route`
+prek hook at commit time and by the `Template-first route (per commit)` CI job
+over every commit a PR adds. State cannot catch this — a hand-edit that happens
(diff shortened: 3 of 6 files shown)


pando-ramet:
CI is red here on a base-branch failure, not on this change: the six failing assertions are `test_e2e.py`'s expected trees missing `ci-ok.yml`, `reference-keywords.json` and `check_gate.py` — the three files !144 added to the render. 256 tests pass; nothing in this PR is implicated, and `ci-ok` is red only because it aggregates them.

The fix is already written and approved in !146 (`8d5eebc`). That PR is one unresolved review thread away from merging, so this branch stays as it is rather than carrying a duplicate of the same commit; once !146 lands I'll merge main in here and CI re-runs against the repaired base.

---

pando-genet (approved):

PR #163, opened 2026-08-12
feat(decision-memory)!: split the preference set — JSON source of truth, plain ordered list as the only injected surface

### LINKED ISSUE #162: preferences: split storage from priming — JSON source of truth, rendered TSV as the only injected surface
Acked by the principal on [decision-memory#14 (comment)](https://github.com/pandoscope/decision-memory/pull/14#issuecomment-5257651566). Storage and priming live in one `preferences.md` today; the metadata suffix costs ~18 tokens per line in a file injected into every session, and the suffix grammar needs bespoke parse/format/strip code in the validator, the guard, and the writer.

**Target format**

`preferences.json` — machine-owned source of truth:

```json
{"rules": [
  {"section": "process",
   "rule": "Prefers machine checks over model checks wherever feasible.",
   "confirmed": 3, "independent": 1, "last": "2026-08-11"}
]}
```

`preferences.txt` — rendered from the JSON, the ONLY injected surface. TSV with one header line, `#` section markers:

```text
confirmed	independent	rule
# process
3	1	Prefers machine checks over model checks wherever feasible.
```

- `last` lives only in the JSON — it matters at update time, never in-session.
- Counters stay injected (grilling cites rules by strength) at ~3 tokens per line instead of ~18. The current 499-token set renders to roughly 300.
- The pair is a declared mirror with a machine check: CI re-renders the JSON and fails on any diff, so the copy cannot drift silently and no model maintains the sync by hand.
- A `rule` string may hold a joined qualifier sentence under the one counter — "one line, one preference" counts preferences, not sentences (ruled 2026-08-11).

**Scope**

- `decision_validator.py`: JSON schema check + renderer replace the suffix grammar (`parse_metadata`/`format_metadata`/`strip_metadata`/`check_metadata_suffix` go away).
- `guards.py`: pref-confirm math validated structurally on before/after JSON instead of unified-diff line pairs; budget measured on the rendered file.
- `preferences_guard.py`: carve-out classification on structural JSON edits; replay-report sha binds to the rendered head file.
- `budget.py`: measures `preferences.txt`.
- `replay.py`: default preference-set path becomes the rendered file.
- `similarity.py`: `preferences_at(«commit»)` falls back to `preferences.md` for pre-migration commits — records pin historical sets; `rule_lines` parses both formats.
- `record.py`: `submit` bumps a JSON field and re-renders; `open`/`check` messages and budget path follow.
- Seeds + `copier.yml` `_skip_if_exists`: `preferences.json` and `preferences.txt` seeded once, store-owned.
- Workflows (`preferences-budget.yml` paths/title), `docs/conventions.md` (format section incl. retiring the "why not JSON" rationale via its own revisit clause), extract-/compact-preferences skills, glossary.
- Migration verb converting a store's `preferences.md` in place; consumer stores stay red with an explicit instruction until they run it — loud, never silent.

Sequenced per the principal's ruling: after decision-memory#5 and decision-memory#14, which both merged.

### LINKED ISSUE #164: extraction: tally rule-vs-rule conflict outcomes against the priority order, propose reorders
Follow-up to #162. Since the ordering ruling (2026-08-14, landed in #163), position in `preferences.json` is priority: earlier rules win when two rules match contradicting solutions. The order is human-owned; this ticket gives it its evidence stream.

**The idea, plainly**

Some decisions are a fight between two rules. One option follows rule A, another follows rule B, and the decider picks one — so one rule won and one lost, with the decider as judge.

The records already remember these contests. Every option lists the rules behind it (`options[].rules_cited`), and the record says which option was chosen. So every contest can be found just by reading the store. Nothing new gets written down.

We count the wins. For each pair of rules: how often did A beat B, how often did B beat A. That is the whole score — a table of counts.

The file's order says who should win: the earlier rule. So the counts are compared to the order. If a rule near the bottom keeps beating a rule above it, the order is probably wrong.

When that happens, the extraction pass writes a note: "these records show X beating Y — consider moving X up." A human looks at it and reorders, or doesn't. Nothing moves a rule by itself — same manner as drift flags today.

Why not Elo: Elo keeps a running score that changes after every game. It depends on the order the games happened, it has to be stored, and it can never be rebuilt exactly. None of that is needed here — the store never forgets a decision, so the count can be redone from zero every time. With 14 rules and a handful of contests, plain counting IS the fair ranking, and it can be checked by hand. (If contest volume ever makes weighting matter, fit Bradley–Terry over the full corpus at read time — still nothing stored.)

A new rule starts at the bottom, the weakest spot. It only moves up when a human moves it — ideally because the count table shows it winning.

Citations are matched to current rules with the same normalized-containment match the counter bumps use, so the two features cannot disagree about which rule a citation names.

**Acceptance**

- `batch` output carries `conflicts` (scope) and `conflict_tally` (corpus), derived, never stored.
- The extract-preferences skill documents the fourth outcome: an order violation gets a reorder proposal file, never a silent reorder.
- Tests: a synthetic corpus with a later-beats-earlier pattern produces the violation report; a conflict-free corpus reports none.

Sequenced after #163 merges and the store converts (the order only exists in the new format).

pando-ramet:
CLOSES #162
CLOSES #164

Implements the format acked on [decision-memory#14 (comment)](https://github.com/pandoscope/decision-memory/pull/14#issuecomment-5257651566), refined by seven later rulings: `preferences.json` becomes the machine-owned source of truth, `preferences.txt` its render and the ONLY file injected into sessions — a plain ordered rule list, one rule per line, nothing else.

**What changed**

- `decision_validator.py`: preference schema (closed key set `rule`/`confirmed`/`independent`/`last`; one-line single-spaced rules with room for a joined qualifier sentence under one counter; unique rule text), deterministic plain-list renderer, canonical serializer. The suffix grammar (`parse_metadata`/`format_metadata`/`strip_metadata`/`check_metadata_suffix`) is gone.
- **The render carries no bookkeeping**: counters and dates live only in the JSON — order already carries the only in-session ranking, and numbers in the render would invite the reader to discount young rules, when a promoted rule is equally binding at zero confirmations. An empty set renders an empty file.
- **Rule order is priority, human-owned**: earlier rules win when two rules match contradicting solutions; new rules append at the end; moving one is a rewrite of the active set and takes the carve-out label — which falls out of the existing change classifier, since a reorder is not a pure addition. Sections (a day-one seed artifact, never decided) are removed entirely.
- **Conflict tally (#164)**: `extraction.py batch` reports every decided rule-vs-rule contest (the chosen option's cited rules beat the declined options' cited rules) plus a corpus-wide win/loss tally per pair of current rules, resolved with the bumps' normalized-containment match. A pair where the later rule out-won the earlier one is flagged `order_violation`; the extract-preferences skill turns that into a reorder *proposal* — a human moves the rule, nothing automatic. Deliberately a recount from the append-only corpus, not a stored Elo-style rating.
- `guards.py`: pref-confirm math validated structurally on before/after JSON; one classifier (`classify_preferences_change`) feeds both the commit guard and the carve-out, so they cannot disagree; the mirror is re-rendered and compared per commit and at head; a missing source is a finding rather than a skipped check.
- `preferences_guard.py`: carve-out classification on structural JSON changes; the replay report's sha binds to the rendered head file.
- `budget.py` / `replay.py` / `similarity.py`: measure and read the rendered file; `rule_lines` reads the one format the render produces.
- `record.py`: `submit` bumps the JSON field and re-renders both files per `pref-confirm` commit; `open` tells the session to inject `preferences.txt`; `check` validates schema, mirror, and budget, and reports a missing source instead of passing silently.
- `render_preferences.py` (new store tool): `render` and `check` — the mirror's update mechanism and its drift alarm.
- Seeds + `copier.yml`: the pair is seeded once and store-owned; docs, skills, and glossary updated.

**Review pass over the branch** found and fixed three defects of one shape — a check that skipped instead of speaking: a vanished `preferences.json` passed `check_corpus` clean (pre-existing on main for the old filename, sharpened by the split); the same hole existed in `record.py check`; and the transitional dual-format parser read a markdown set's prose as rules, which would have inflated coverage scores and manufactured false-cold flags. The sweep that followed removed every remaining reference to the earlier format after judging each on whether it still had a job — none did. Stale test fixtures surfaced honestly in the process: two size-invariance katas had been passing on a diluted score, and a verbatim rule quote now measures exactly 1.0 coverage.

**Conversion for the existing store**

No one-shot migrate tooling ships — the format ends here, and the single live store writes its `preferences.json` (with an explicit priority ordering, now that order means something) in its own PR; the schema, mirror, and budget guards verify the result mechanically. Until that lands, guards and recorder fail loudly that the set is missing — never a silent green.

**Notes for review**

- Overlaps #154 (`docs/conventions.md`, same section) — whichever merges second takes a small rebase.
- Found but not fixed (pre-existing, fails on an unmodified tree too): `tests/test_record_session.py` only passes when `tests/test_record_core.py` runs first, because the fixture store copies `record.py` without `record_core.py` and relies on the earlier test's `sys.modules` cache.
- Verified: subtemplate store suite green; root pytest 267 green; prek clean on every commit.


decision-memory/.github/guards/decision_validator.py
```diff
@@ -84,102 +85,151 @@
...
+    so callers can decide how far to trust it.
+    """
+    try:
+        data = json.loads(text)
+    except json.JSONDecodeError as exc:
+        return None, [f"not valid JSON: {exc}"]
+    return data, validate_preferences(data)
+
+
+def _validate_rule(index: int, rule: object) -> list[str]:
```
pando-genet, line 129: Why not use pydantic?

decision-memory/.github/store/render_preferences.py
```diff
@@ -0,0 +1,236 @@
...
+
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+sys.path.insert(
+    0,
+    os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))), "guards"),
+)
+
+import decision_validator  # noqa: E402  (path bootstrap above)
+
+LEGACY_FILENAME = "preferences.md"
```
pando-genet: We don't need this, we'll get rid of the file in this PR

pando-genet (changes requested):

decision-memory/.github/guards/decision_validator.py
pando-ramet: No commit: the vendored guard stack is stdlib-only by contract — its docstring promises the copy keeps working even if the template repo disappears, CI runs it on a bare setup-python with no install step, and the recorder imports this same file inside arbitrary store clones. Pydantic would add that dependency to every store's CI and every recording session, and it wouldn't remove the checks that carry the weight here: the per-field shapes are five lines, while the real rules are cross-field and cross-rule (closed key order, unique rule text, contiguous sections, independent ≤ confirmed, and the render-equality mirror), which stay hand-written either way. That's the standing "rejects a new dependency unless it removes a whole class of maintenance" line — here it removes almost nothing. If the schema ever grows real nesting, that's the point to revisit.

---

decision-memory/.github/store/render_preferences.py
pando-ramet: Dropped — the migrate verb, the legacy parser, and this constant are gone; `render_preferences.py` is down to `render`/`check`. The store converts its rules by hand once, and the schema + mirror + budget guards verify the result mechanically. The pre-split and leftover-file checks in the guards stay, so a store still carrying `preferences.md` fails loudly with the instruction to convert (kept `similarity.py`'s read-side fallback too: records pin `preference_set.commit`s from before the split, and the file at those historical commits is markdown forever).

https://github.com/pandoscope/agentic-engineering-template/commit/73554fab50d27008c15b312be4c8600f4cba0964

---

pando-genet (approved):

PR #165, opened 2026-08-15
feat(store): required-nullable doc field — promotion-doc provenance per preference rule

pando-ramet:
ADVANCES pandoscope/decision-memory#20

**Reworked onto the post-#163 format** (the original sidecar approach collided with `preferences.json` becoming the source of truth — see the earlier comment; the branch was rebuilt from current main, no sidecar remnants).

## What

Every `preferences.json` rule entry now carries a fifth key, **`doc`** — the promotion-doc provenance:

- **Required-but-nullable**: `doc` is either the http(s) URL of the doc that promoted the rule (typically its `proposals/` file) or `null` as the explicit "no promotion doc exists" declaration. A missing key fails the closed-key-set check, so absence is always observable — "no doc" and "forgot to write it" can never look the same.
- `decision_validator.py`: `DOC_KEY` added to `RULE_KEYS`; `_validate_rule` accepts null or an http(s) URL and rejects everything else with a field-naming error. `render_preferences` is untouched — the injected `preferences.txt` stays a plain rule list, and a test now pins that the render never emits the doc.
- Written by the promoting human in the promoting commit; the mirror/carve-out/budget guards all apply unchanged since `doc` lives inside the existing source of truth — no second file, nothing new to drift.
- `docs/conventions.md`: schema example and closed-set bullet updated; new `doc` bullet under the active-preference-set section.

## Tests

Test-first: `PreferenceDocTests` in the subtemplate store suite (null passes, URL passes, missing key rejected, non-URL/bool/int rejected, render never emits) — red at 8 errors before the implementation. Both rule factories gained `doc=None`. Subtemplate suite 177 green; root pytest 271 green (copier e2e renders included, so seeds and the recorder's counter bumps survive the change).

## Consumers

The single live store backfills `doc` values in its conversion-era PR (pandoscope/decision-memory!21 is retired in favor of that); the grilling skill fills `preferenceDocs` from `preferences.json` at injection time (pandoscope/skills#134).

### DIFF
--- a/decision-memory/.github/guards/decision_validator.py
+++ b/decision-memory/.github/guards/decision_validator.py
@@ -98,6 +98,10 @@ PREFERENCES_RENDERED = "preferences.txt"
 COUNTER_KEY = "confirmed"
 INDEPENDENT_KEY = "independent"
 DATE_KEY = "last"
+# Promotion-doc provenance, required-but-nullable: null is the explicit
+# "no promotion doc exists" declaration, so a missing key is always a
+# defect rather than an ambiguity. The render never emits it.
+DOC_KEY = "doc"
 
 # Exactly these, on every rule. A closed set is what keeps a rule from
 # gaining keys nothing reads: adding one is a deliberate change here,
@@ -109,7 +113,7 @@ DATE_KEY = "last"
 # append at the end; moving one is a rewrite of the active set and
 # gated like any other. Nothing here enforces a particular order — the
 # order IS the ruling.
-RULE_KEYS = ("rule", COUNTER_KEY, INDEPENDENT_KEY, DATE_KEY)
+RULE_KEYS = ("rule", COUNTER_KEY, INDEPENDENT_KEY, DATE_KEY, DOC_KEY)
 
 
 def parse_preferences(text: str) -> tuple[dict | None, list[str]]:
@@ -155,6 +159,14 @@ def _validate_rule(index: int, rule: object) -> list[str]:
         )
     if not isinstance(rule[DATE_KEY], str) or not DATE_RE.fullmatch(rule[DATE_KEY]):
         errors.append(f"{where}: {DATE_KEY}={rule[DATE_KEY]!r} is not YYYY-MM-DD")
+    doc = rule[DOC_KEY]
+    if doc is not None and (
+        not isinstance(doc, str) or not doc.startswith(("http://", "https://"))
+    ):
+        errors.append(
+            f"{where}: {DOC_KEY}={doc!r} must be null (no promotion doc "
+            "exists) or the http(s) URL of the doc that promoted the rule"
+        )
     return errors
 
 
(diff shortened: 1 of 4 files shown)


pando-ramet:
Incompatible with #163 as merged — holding this as draft rather than iterating blind.

This PR claims `preferences.json` as a provenance *sidecar* (rule text → doc URL) and validates it against `preferences.md`'s counter-suffixed bullets. #163 makes `preferences.json` the machine-owned **source of truth** (closed key set `rule`/`confirmed`/`independent`/`last`, rendered to `preferences.txt`) and deletes the suffix grammar this guard parses. Same filename, two owners, and the parsing substrate is gone.

The provenance goal survives; the shape should follow the new format: an optional `doc` key added to the preference schema's closed key set (per-rule, machine-owned, travels with the rule through reorders and rewording), with the renderer continuing to exclude it from `preferences.txt` — no separate file, no drift pair. Consumers (grilling's `preferenceDocs`, pandoscope/decision-memory#20) read it from the JSON.

Will rework on that basis once the design is acked on decision-memory#20.

---

pando-genet (dismissed):

pando-genet (dismissed):

pando-genet (changes requested):
Shorten CLAUDE.md addition

template/CLAUDE.md.jinja
```diff
@@ -28,6 +28,12 @@ API budget per-account, shared by all sessions.
 New non-obvious findings — tools, environment, failure modes — go in
 this file, tersely.
 
+- Managed-environment clones do not all get `prek install` run: commits
+  in a clone without `.git/hooks/pre-commit` skip every hook silently
+  and defects surface only in CI. A commit that prints no hook ladder
+  is the tell; run `prek install` (or `uv run prek install`) once per
+  clone before the first commit.
```
pando-genet: Can this be mechanically enforced? It would save token budget here and be more reliable

pando-genet (changes requested):

pando-ramet:
Re the review comment on the CLAUDE.md learning ("can this be mechanically enforced?"): yes — and it turned out it already is. `scripts/ensure-prek.sh` is the mechanism (SessionStart hook, multi-root aware since #139: run from a parent dir it installs into every `*/.git` clone). The miss that prompted the learning was harness-side: in a managed multi-root session the project dir is the parent, per-repo SessionStart hooks never fire, and nothing org-level invokes the script — so this clone commited hookless while its siblings had hooks.

Acted on it: the learning lines are **removed** from this PR (last two commits — source + reapply stamp; the token budget goes back to zero) and the actual enforcement gap is filed as pandoscope/meta#84: the org-level session startup invokes each clone's `ensure-prek.sh`, which its multi-root branch already makes a one-call fix.

(Replying at PR level because the review-thread API is currently rate-limited for this session; happy to resolve the inline thread once it recovers.)

---

pando-genet (approved):

Issue #199, opened 2026-09-01
CLAUDE.md drifts per-repo — detect drift and consolidate at template level


pando-ramet:
Ruling on [meta!114](https://github.com/pandoscope/meta/pull/114#discussion_r3905879093): `CLAUDE.md` should never be changed in a consumer repo, only at template level — yet consumer copies have drifted. meta's copy carries a session-maintained "Learnings" section (hook channels, compaction, org-enumeration proxy policy, Forge Budget wording differences), and sessions keep appending to it because the file itself invites terse findings.

Asks:

- **Drift check**: a guard (CI or scheduled) that compares each consumer's `CLAUDE.md` against the rendered template output and flags divergence, the same way byte-pinned files are checked.
- **Consolidation**: fold the useful drifted content (e.g. meta's Learnings entries) back into the template render — or move repo-specific learnings to a file the template does not own — then restamp so every consumer converges.
- The template's stamped `CLAUDE.md` should stop inviting in-place edits (the "Learnings … go in this file" instruction) if the file is template-owned.

One entry was dropped from [meta!114](https://github.com/pandoscope/meta/pull/114) per the ruling and belongs in whatever consolidated home this ticket decides: CCW env vars refresh per turn, not per container — an edited environment value (e.g. PUSH_BLOCKLIST) reaches a running remote session on its next turn, no restart needed (observed twice, 2026-08-31).

PR #213, opened 2026-09-02
fix(ci): the drift check's recopy never waits on a prompt

pando-ramet:
ADVANCES #199.

Measured on pandoscope/disambiguate#82 ([job](https://github.com/pandoscope/disambiguate/actions/runs/33652347782/job/100322253112)): the stamped lint workflow's "vendored files match the stamped template" job runs `copier recopy`, which asks `Overwrite CLAUDE.md? (Y/n)` wherever CLAUDE.md diverged — and a runner has no terminal, so the job dies with "Interactive session required: Consider using `--overwrite`" before the diff ever runs. The check never reached its verdict.

`--overwrite` gets past the prompt so the diff runs. Per the review ruling, CLAUDE.md is judged like every other stamped file — no restore exempts it; a repo whose CLAUDE.md drifted goes red, loudly. (The earlier exemption in this workflow and the "Learnings go in this file" invitation in the stamped CLAUDE.md contradict that ruling; the invitation is #199's consolidation work.) Red-first test in `tests/test_ci_gate.py`. The root repo carries no `lint.yml`, so no restamp commit.

https://claude.ai/code/session_01DC3EFouHkiuQcB7kRN5gpa

### DIFF
--- a/template/{% if agentic_forge == 'github' %}.github{% endif %}/workflows/lint.yml.jinja
+++ b/template/{% if agentic_forge == 'github' %}.github{% endif %}/workflows/lint.yml.jinja
@@ -58,9 +58,13 @@ jobs:
 
           # Seeded (_skip_if_exists) files already exist, so recopy
           # leaves them alone — they stay exempt structurally.
+          # --overwrite: recopy prompts "Overwrite CLAUDE.md?" wherever
+          # that file carries the local Learnings it invites, and a
+          # runner has no terminal to answer — the restore below is
+          # what keeps that content exempt, not the prompt.
           uvx --with copier-template-extensions copier recopy \
             --answers-file "$answers_file" \
-            --defaults --trust --skip-tasks --vcs-ref "$stamped"
+            --defaults --trust --skip-tasks --overwrite --vcs-ref "$stamped"
 
           # The recopy re-delivers every shared glossary term, including
           # the ones this repo's stamp pruned as unlinked (#67). Prune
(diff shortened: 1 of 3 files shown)


template/{% if agentic_forge == 'github' %}.github{% endif %}/workflows/lint.yml.jinja
```diff
@@ -58,9 +58,13 @@ jobs:
...
           # Seeded (_skip_if_exists) files already exist, so recopy
           # leaves them alone — they stay exempt structurally.
+          # --overwrite: recopy prompts "Overwrite CLAUDE.md?" wherever
+          # that file carries the local Learnings it invites, and a
+          # runner has no terminal to answer — the restore below is
+          # what keeps that content exempt, not the prompt.
           uvx --with copier-template-extensions copier recopy \
             --answers-file "$answers_file" \
-            --defaults --trust --skip-tasks --vcs-ref "$stamped"
+            --defaults --trust --skip-tasks --overwrite --vcs-ref "$stamped"
```
pando-genet, line 68: We want to fail loudly, not drift silently (unless the need explicitly arises in a concrete scenario,  then we can make an exemption).
Claude.md should not drift in any repo

pando-genet (changes requested):

template/{% if agentic_forge == 'github' %}.github{% endif %}/workflows/lint.yml.jinja
pando-ramet: Done in 706d92d: the CLAUDE.md restore is gone, so the diff judges CLAUDE.md like every other stamped file and a drifted repo goes red loudly. `--overwrite` stays only so a runner with no terminal gets past copier's prompt and reaches that verdict. Test asserts both. Noted the contradiction that remains template-side: the stamped CLAUDE.md still says "Learnings go in this file" — that invitation is #199's consolidation work.

---

